PR #1, opened 2020-02-10
Initial scoped locals implementation

theRealAph:
Here it is. Still no Structured Concurrency support, though.

## Progress
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed


theRealAph:
This failed because there is no Bug ID, and no reviewer. But the other Loom commits don't have these. What do you want me to do?

pron:
Changed jcheck configuration to allow this in 07e3880.

PR #16, opened 2020-09-15
8246039: SSLSocket HandshakeCompletedListeners are run on virtual threads

carterkozak:
Previously each `SSLSocket.addHandshakeCompletedListener(HandshakeCompletedListener)`
resulted in a new OS thread for each handshake which causes spikes in
resource utilization and bottlenecks in high throughput systems.

Virtual threads should reduce the overhead while preserving behavior to
the extent possible.

Reported to openjdk security-dev here:
https://mail.openjdk.java.net/pipermail/security-dev/2020-July/022220.html

---------
### Progress

### Reviewers
 * [Bradford Wetmore](https://openjdk.java.net/census#wetmore) (@bradfordwetmore - no project role)

### Download
`$ git fetch https://git.openjdk.java.net/loom pull/16/head:pull/16`
`$ git checkout pull/16`

### DIFF
--- a/src/java.base/share/classes/sun/security/ssl/TransportContext.java
+++ b/src/java.base/share/classes/sun/security/ssl/TransportContext.java
@@ -637,13 +637,11 @@ final class TransportContext implements ConnectionContext {
                 !sslConfig.handshakeListeners.isEmpty()) {
             HandshakeCompletedEvent hce =
                 new HandshakeCompletedEvent((SSLSocket)transport, conSession);
-            Thread thread = new Thread(
-                null,
-                new NotifyHandshake(sslConfig.handshakeListeners, hce),
-                "HandshakeCompletedNotify-Thread",
-                0,
-                false);
-            thread.start();
+            Thread.builder()
+                    .virtual()
+                    .name("HandshakeCompletedNotify-Thread")
+                    .task(new NotifyHandshake(sslConfig.handshakeListeners, hce))
+                    .start();
         }
 
         return HandshakeStatus.FINISHED;


carterkozak:
/signed

However my OCA doesn't have github information. Should I submit an updated OCA with my github account info?

AlanBateman:
Thanks for identifying this potential use-case.

I found the discussion from Jan 2013 on security-dev about this issue but I was unable to find a conclusion or the rational for why the listeners weren't called directly (which would of course would stall finishHashshake). So I think we should re-start that discussion and get a conclusion on whether this will be changed or not. Minimally we need javax.net.ssl.SSLSocket clarified to set expectations on the execution context and whether it runs asynchronously from the finish (the session may be invalidated before or while the listener runs).

Assuming security-dev wants to continue to create thread per handshake notification then your patch make sense, just need to drop inheritThreadLocals (the final parameter Thread constructor in TransportContext is false, meaning that inheritable thread locals are not inherited in the existing code).

carterkozak:
> I think we should re-start that discussion and get a conclusion on whether this will be changed or not.

I completely agree that running listeners on the completion thread would be ideal and allow the most flexibility for consumers, but I'm worried that changing the behavior between java releases without a new API (perhaps an overload which additionally takes an Executor instance) would cause too much friction for existing consumers. Listeners which expect to execute long running operations would potentially cause performance regressions when taking a java release including the change, which I'd imagine would make it a non-starter. Loom virtual threads provide the ability to preserve existing behavior without the overhead incurred by OS threads.
I'd be happy to start a thread to discuss a mechanism to listen for handshake completion without dispatching to additional threads (virtual or otherwise), but I think that would be a separate, parallel, change. What do you think?

> just need to drop inheritThreadLocals (the final parameter Thread constructor in TransportContext is false, meaning that inheritable thread locals are not inherited in the existing code)

Good catch, I got my wires crossed! Fixed.

bradfordwetmore:
From a coding point of view, if Xuelei doesn't have a further suggestion, using virtual threads like you have suggested seems to be a good solution.  I'm ok with this proposed change for Project Loom.  Loom has a lot of promise for issues like this.  Looking through the Thread builder source, it looks like right options are captured.

However, let's give Xuelei a bit of time to think of other options before integration.

bradfordwetmore (approved):
Please wait for further consideration from Xuelei, but the changes seem good to me.

AlanBateman:
https://mail.openjdk.java.net/pipermail/security-dev/2020-September/022482.html

There is musing on possible alternatives to fix the issue but I think Brad and Xuelei are okay with changing it to start a virtual thread per notification for now and see how it goes. So I'll merge this PR.

uhla:
Hi @uhla, thanks for making a comment in an OpenJDK project!

All comments and discussions in the OpenJDK Community must be made available under the OpenJDK [Terms of Use](https://openjdk.java.net/legal/tou/). If you already are an OpenJDK [Author](https://openjdk.java.net/bylaws#author), [Committer](https://openjdk.java.net/bylaws#committer) or [Reviewer](https://openjdk.java.net/bylaws#reviewer), please click [here](https://bugs.openjdk.java.net/secure/CreateIssue.jspa?pid=11300&issuetype=1) to open a new issue so that we can record that fact. Please Use "Add GitHub user uhla" for the summary.

If you are not an OpenJDK Author, Committer or Reviewer, simply check the box below to accept the OpenJDK Terms of Use for your comments.

 - [ ]  I agree to the [OpenJDK Terms of Use](https://openjdk.java.net/legal/tou/) for all comments I make in a project in the [OpenJDK GitHub organization](https://github.com/openjdk).

Your comment will be automatically restored once you have accepted the OpenJDK [Terms of Use](https://openjdk.java.net/legal/tou/).

uhla:
Hi @uhla, thanks for making a comment in an OpenJDK project!

All comments and discussions in the OpenJDK Community must be made available under the OpenJDK [Terms of Use](https://openjdk.java.net/legal/tou/). If you already are an OpenJDK [Author](https://openjdk.java.net/bylaws#author), [Committer](https://openjdk.java.net/bylaws#committer) or [Reviewer](https://openjdk.java.net/bylaws#reviewer), please click [here](https://bugs.openjdk.java.net/secure/CreateIssue.jspa?pid=11300&issuetype=1) to open a new issue so that we can record that fact. Please Use "Add GitHub user uhla" for the summary.

If you are not an OpenJDK Author, Committer or Reviewer, simply check the box below to accept the OpenJDK Terms of Use for your comments.

 - [ ]  I agree to the [OpenJDK Terms of Use](https://openjdk.java.net/legal/tou/) for all comments I make in a project in the [OpenJDK GitHub organization](https://github.com/openjdk).

Your comment will be automatically restored once you have accepted the OpenJDK [Terms of Use](https://openjdk.java.net/legal/tou/).

PR #18, opened 2020-10-13
Cleanup for v1

theRealAph:
This is the Big Diff that removes scope locals, Lifetimes, etc. I've left the virtual machine support in for now. I'm also going to create a separate branch with the simple virtual ThreadLocal support + Lifetimes, and we'll see what that looks like.

---------
### Progress

### Download
`$ git fetch https://git.openjdk.java.net/loom pull/18/head:pull/18`
`$ git checkout pull/18`


AlanBateman:
I guess we can close this.

PR #19, opened 2020-10-13
Slightly simpler way of waiting for a specific time

kabutz:
Instead of writing the logic for calling System.nanoTime() etc. ourselves, could we perhaps use awaitNanos() instead?

---------
### Progress

### Download
`$ git fetch https://git.openjdk.java.net/loom pull/19/head:pull/19`
`$ git checkout pull/19`


AlanBateman:
Yes, better to use awaitNanos rather than await here.

PR #22, opened 2020-12-01
fix typo

hepin1989:


src/hotspot/share/runtime/continuation.hpp
```diff
@@ -88,7 +88,7 @@ class Continuation : AllStatic {
 
   static void notify_deopt(JavaThread* thread, intptr_t* sp);
 
-  static oop  get_continutation_for_frame(JavaThread* thread, const frame& f);
+  static oop  get_continuation_for_frame(JavaThread* thread, const frame& f);
```
hepin1989, line 91: get_continutation_for_frame

hepin1989: @pron I think this is a typo

pron:
I've fixed this typo, which actually wasn't trivial due to C++'s overload resolution and a one-argument constructor that should have been `explicit` but wasn't. Thank you for noticing it.

PR #38, opened 2021-05-04
Loom support for RedefineClasses

coleenp:
Use _marking_cycle to determine whether a Method* is on the heap.
This patch includes the patches for https://github.com/openjdk/jdk/pull/3852 and https://git.openjdk.java.net/jdk/pull/3861 which are going to be pushed to mainline.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/38/head:pull/38` \
`$ git checkout pull/38`

`$ git checkout pull/38` \
`$ git pull https://git.openjdk.java.net/loom pull/38/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 38`

`$ git pr show -t 38`

</details>

<a href="https://git.openjdk.java.net/loom/pull/38.diff">https://git.openjdk.java.net/loom/pull/38.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/oops/instanceStackChunkKlass.inline.hpp
+++ b/src/hotspot/share/oops/instanceStackChunkKlass.inline.hpp
@@ -793,11 +793,8 @@ void InstanceStackChunkKlass::run_nmethod_entry_barrier_if_needed(const StackChu
   CodeBlob* cb = f.cb();
   if ((mixed && cb == nullptr) || !cb->is_nmethod()) {
     // Mark interpreted frames for marking_cycle
-    // if jvmtiExport::has_redefined_a_class() ???
     assert(f.is_interpreted(), "what else?");
     Method* im = f.to_frame().interpreter_frame_method();
-    ResourceMark rm;
-    tty->print_cr("marking method %s",im->name_and_sig_as_C_string());
     im->record_marking_cycle();
   } else {
     nmethod* nm = cb->as_nmethod();
--- a/src/hotspot/share/oops/instanceStackChunkKlass.inline.hpp
+++ b/src/hotspot/share/oops/instanceStackChunkKlass.inline.hpp
@@ -791,10 +791,19 @@ inline BitMap::idx_t InstanceStackChunkKlass::bit_offset(int stack_size_in_words
 template <bool mixed>
 void InstanceStackChunkKlass::run_nmethod_entry_barrier_if_needed(const StackChunkFrameStream<mixed>& f) {
   CodeBlob* cb = f.cb();
-  if ((mixed && cb == nullptr) || !cb->is_nmethod()) return;
-  nmethod* nm = cb->as_nmethod();
-  if (BarrierSet::barrier_set()->barrier_set_nmethod()->is_armed(nm)) {
-    nm->run_nmethod_entry_barrier();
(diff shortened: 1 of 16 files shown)


src/hotspot/share/oops/instanceStackChunkKlass.inline.hpp
```diff
@@ -791,10 +791,16 @@ inline BitMap::idx_t InstanceStackChunkKlass::bit_offset(int stack_size_in_words
 template <bool mixed>
 void InstanceStackChunkKlass::run_nmethod_entry_barrier_if_needed(const StackChunkFrameStream<mixed>& f) {
   CodeBlob* cb = f.cb();
-  if ((mixed && cb == nullptr) || !cb->is_nmethod()) return;
-  nmethod* nm = cb->as_nmethod();
-  if (BarrierSet::barrier_set()->barrier_set_nmethod()->is_armed(nm)) {
-    nm->run_nmethod_entry_barrier();
+  if ((mixed && cb == nullptr) || !cb->is_nmethod()) {
```
pron, line 794: If `mixed` is false, then there are no interpreted frames. The condition should be `if (f.is_interpreted())` (which short-circuits to false if `mixed` is false).

test/hotspot/jtreg/ProblemList.txt
```diff
@@ -167,7 +167,7 @@ compiler/vectorization/TestMacroLogicVector.java                              82
 compiler/whitebox/ForceNMethodSweepTest.java                                  8265181 generic-all
 
 runtime/clinit/ClassInitBarrier.java                                          8254936 generic-all
-runtime/vthread/RedefineClass.java                                            8253378 generic-all
+# runtime/vthread/RedefineClass.java                                            8253378 generic-all
```
coleenp, line 170: I should delete this line.

src/hotspot/share/oops/instanceStackChunkKlass.inline.hpp
coleenp: Ok that's much better.  Thanks.  If (!f.is_interpreted) then cb will always be an nmethod? Or do I need to check for that in the else case?

coleenp: see revised version.

pron: There's one other possibility: a safepoint stub (in the case of forced preemption). But `StackChunkFrameStream` also has `is_compiled` and `is_stub`. Exactly one of those three would be true for each frame. If `is_compiled`, then you have an nmethod.

coleenp: I changed this so that it checks if it's interpreted, and then in the else check if cb->is_nmethod() to exclude the stub case.  Does it look right?

pron: Yep. Looks good.

coleenp: Thanks!

sspitsyn:
It looks okay too me.

PR #43, opened 2021-05-14
Fix preemption to not try to allocate heap but not GC if it fails, so…

coleenp:
… that it can be called in a safepoint.
Sorry for the whitespace changes.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/43/head:pull/43` \
`$ git checkout pull/43`

`$ git checkout pull/43` \
`$ git pull https://git.openjdk.java.net/loom pull/43/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 43`

`$ git pr show -t 43`

</details>

<a href="https://git.openjdk.java.net/loom/pull/43.diff">https://git.openjdk.java.net/loom/pull/43.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/g1/g1CollectedHeap.cpp
+++ b/src/hotspot/share/gc/g1/g1CollectedHeap.cpp
@@ -404,6 +404,15 @@ G1CollectedHeap::mem_allocate(size_t word_size,
   return attempt_allocation(word_size, word_size, &dummy);
 }
 
+// This function is called with the Heap_lock held.  It attempts to allocate memory
+// but if it fails, just returns NULL.  It doesn't safepoint.
+HeapWord* G1CollectedHeap::try_mem_allocate(size_t word_size) {
+  NoSafepointVerifier nsv;
+  assert_locked_or_safepoint_weak(Heap_lock);
+  assert (!is_humongous(word_size), "Humongous is not supported");
+  return _allocator->attempt_allocation_locked(word_size);
+}
+
 HeapWord* G1CollectedHeap::attempt_allocation_slow(size_t word_size) {
   ResourceMark rm; // For retrieving the thread names in log messages.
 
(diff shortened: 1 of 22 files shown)


coleenp:
I meant to make this a draft.

src/hotspot/share/gc/g1/g1CollectedHeap.cpp
```diff
@@ -404,6 +404,15 @@ G1CollectedHeap::mem_allocate(size_t word_size,
...
 
+// This function is called with the Heap_lock held.  It attempts to allocate memory
+// but if it fails, just returns NULL.  It doesn't safepoint.
+HeapWord* G1CollectedHeap::try_mem_allocate(size_t word_size) {
+  NoSafepointVerifier nsv;
+  assert_locked_or_safepoint_weak(Heap_lock);
+  assert (!is_humongous(word_size), "Humongous is not supported");
+  return _allocator->attempt_allocation_locked(word_size);
+}
+
```
coleenp, line 415: I get a card table failure so this might need a lot more work (and expertise!)

robehn:
Hi Coleen,

A handshake can take a long time, e.g. suppose you have 100 threads on 4 core machine with 100HZ, in theory you would get 4 slices per second, in practice not near that much, which means Heap_lock could be locked for half a second or more.
If you do more than two preemptive handshakes per second the Heap_lock will be always locked in such case.
Also this means the thread doing the synchronous handshake cannot do any other work, I would guess this is scheduling thread? And we probably want to schedule in the mean-time?

The Z case is more complicated since you must not either induce an allocation stale, which means we could be waiting for a second thread to safepoint for us. But we are unsafe so the safepoint will never happen. So the NoSafepointVerifier nsv; in the Z case does not help AFAICT.

If you do the preemptive in an asynchronous handshake:
- The requesting thread can continue with work (scheduling?).
- You could safepoint, and I _think_ allocation stale is fine.
- The preempted thread will call-back to scheduler, so we do not need another notification?

coleenp:
Ok, this change is bad.

PR #60, opened 2021-09-06
Ensure vthread has connected to server

miao-zheng:
I have run the test case of InterruptHttp.java in fast debug version of jdk with options:
-XX:+UseG1GC  -XX:+ScavengeALot -XX:ScavengeALotInterval=200 -XX:-GCALotAtAllSafepoints

And the test case failed, The output is "Execution failed: `main' threw exception: java.lang.RuntimeException: Expected 1 HTTP connection"

The reason is Thread.sleep(1000) can not ensure that vthread has connected to server when using some options and fastdebug version of jdk.

So, I fix this to ensure vthread has connected to server.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/60/head:pull/60` \
`$ git checkout pull/60`

`$ git checkout pull/60` \
`$ git pull https://git.openjdk.java.net/loom pull/60/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 60`

`$ git pr show -t 60`

</details>

<a href="https://git.openjdk.java.net/loom/pull/60.diff">https://git.openjdk.java.net/loom/pull/60.diff</a>

</details>

### DIFF
--- a/test/jdk/java/lang/Thread/virtual/InterruptHttp.java
+++ b/test/jdk/java/lang/Thread/virtual/InterruptHttp.java
@@ -55,7 +55,9 @@ public class InterruptHttp {
             });
 
             // give time for thread to block waiting for HTTP server
-            Thread.sleep(1000);
+            while (server.connectionCount() == 0) {
+                Thread.sleep(100);
+            }
             thread.interrupt();
             thread.join();
 


test/jdk/java/lang/Thread/virtual/InterruptHttp.java
```diff
@@ -55,7 +55,9 @@ public static void main(String[] args) throws Exception {
             });
 
             // give time for thread to block waiting for HTTP server
-            Thread.sleep(1000);
+            while (thread.getState() != Thread.State.WAITING) {
+                Thread.sleep(1000);
+            }
```
AlanBateman, line 60: Thanks for this. You are right that that the test is problematic. Can you try the following instead of polling the thread state? This will wait until a connection is established so that the read is interrupted.

```
while (server.connectionCount() == 0) {
    Thread.sleep(100);
}
```

miao-zheng: Thanks for your review.
Do I need remove this? I think the following is duplicated with the check of "server.connectionCount() == 0"
```
    // server should have accepted one connection
    if (server.connectionCount() != 1)
        throw new RuntimeException("Expected 1 HTTP connection");
```

AlanBateman: We should keep it to checks that only one connection is established. The details on this are in sun.net.www.http.HttpClient where it makes a decision on whether to retry the HTTP request when interrupted. In this case, it should not retry.

miao-zheng: Got it. Thanks for your answer.
I have changed this PR on your suggestion.

AlanBateman (approved):

PR #65, opened 2021-09-24
Make Loom buildable on x86_32

shipilev:
This is WIP to make Loom workspace at least buildable on x86_32. Since jlink/jmod run build JDK, the JDK itself has to start well, without touching Loom code.


shipilev:
All right, I think I nailed most of the generic x86_32 bugs here, and it builds and runs non-Loom code fine. 

Here it is building Linux x86 cleanly in GHA:
  https://github.com/shipilev/loom/runs/3801890849
  https://github.com/shipilev/loom/runs/3801890884

Please take a look @pron, @AlanBateman.

shipilev:
Cheers.

/integrate

PR #66, opened 2021-09-27
Stop compiling old oopMapStubGenerator

shipilev:
This code is currently not used, and it might not be used in future too. There is no way to test if it works, currently. It is the major source of build failure when compiling x86_32. So, it would be better to `#if 0` it, or even remove it wholesale.

Additional testing:
 - [x] Linux x86_64 `tier1_loom`

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/66/head:pull/66` \
`$ git checkout pull/66`

`$ git checkout pull/66` \
`$ git pull https://git.openjdk.java.net/loom pull/66/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 66`

`$ git pr show -t 66`

</details>

<a href="https://git.openjdk.java.net/loom/pull/66.diff">https://git.openjdk.java.net/loom/pull/66.diff</a>

</details>


pron:
Thank you! You can `/integrate`.

src/hotspot/share/compiler/oopMapStubGenerator.hpp
```diff
@@ -51,3 +54,7 @@ class OopMapStubGenerator {
 };
 
 #endif
+
+#else
+#error This code is not used
+#endif
```
AlanBateman, line 60: Can you check Windows, I suspect it might have to be excluded.

shipilev: Um, Windows builds fail here:
 https://github.com/shipilev/loom/runs/3716995428?check_suite_focus=true

Which implies `oopMapStubGenerator_x86.cpp` is still compiled? What. Let me see...

shipilev: See #68.

PR #67, opened 2021-09-27
Fix Minimal VM build failures

shipilev:
Minimal VM configs have no C2 and no JVMTI. New Loom code fails the Minimal VM builds because of this. See for example GHA runs: https://github.com/shipilev/loom/runs/3717293969?check_suite_focus=true

C2 needs just a little bit of protection when reaching to `DerivedPointersTable`.

JVMTI needs a bit more work. `jvmtiExport` and `jvmtiThreadState` headers are usually included even with JVMTI turned off, but `.cpp` would not be compiled without JVMTI. Therefore, non-trivial implementations should go into `.cpp`. Plus, some of the paths that call the actual methods declared but not implemented without JVMTI should be protected with `INCLUDE_JVMTI`.

Additional testing:
 - [x] Linux x86_64 server, `tier1_loom` still passes (includes JVMTI tests)
 - [x] Linux x86_64 minimal now builds, runs `tier1_loom` (cannot pass JVMTI tests without JVMTI)

---------
### Progress

### Reviewers
 * [Serguei Spitsyn](https://openjdk.java.net/census#sspitsyn) (@sspitsyn - Committer) ⚠️ Review applies to 9ba55eda41cb018ba34f8a791fa34a4b0517d318
 * [Chris Plummer](https://openjdk.java.net/census#cjplummer) (@plummercj - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/67/head:pull/67` \
`$ git checkout pull/67`

`$ git checkout pull/67` \
`$ git pull https://git.openjdk.java.net/loom pull/67/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 67`

`$ git pr show -t 67`

</details>

<a href="https://git.openjdk.java.net/loom/pull/67.diff">https://git.openjdk.java.net/loom/pull/67.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/prims/jvm.cpp
+++ b/src/hotspot/share/prims/jvm.cpp
@@ -3890,12 +3890,14 @@ JVM_END
 JVM_ENTRY(void, JVM_VirtualThreadMountBegin(JNIEnv* env, jobject vthread, jboolean first_mount))
 #if INCLUDE_JVMTI
   JvmtiVTMTDisabler::start_VTMT(vthread, 0);
+#else
+  fatal("Should only be called with JVMTI enabled");
 #endif
 JVM_END
 
 JVM_ENTRY(void, JVM_VirtualThreadMountEnd(JNIEnv* env, jobject vthread, jboolean first_mount))
-  assert(thread->is_in_VTMT(), "VTMT sanity check");
 #if INCLUDE_JVMTI
+  assert(thread->is_in_VTMT(), "VTMT sanity check");
   JvmtiVTMTDisabler::finish_VTMT(vthread, 0);
   oop vt = JNIHandles::resolve(vthread);
 
@@ -3916,6 +3918,8 @@ JVM_ENTRY(void, JVM_VirtualThreadMountEnd(JNIEnv* env, jobject vthread, jboolean
   if (JvmtiExport::should_post_vthread_mount()) {
     JvmtiExport::post_vthread_mount(vthread);
   }
+#else
+  fatal("Should only be called with JVMTI enabled");
 #endif
 JVM_END
 
@@ -3943,12 +3947,16 @@ JVM_ENTRY(void, JVM_VirtualThreadUnmountBegin(JNIEnv* env, jobject vthread, jboo
 
   assert(!thread->is_in_VTMT(), "VTMT sanity check");
   JvmtiVTMTDisabler::start_VTMT(vthread, 1);
+#else
+  fatal("Should only be called with JVMTI enabled");
 #endif
 JVM_END
 
 JVM_ENTRY(void, JVM_VirtualThreadUnmountEnd(JNIEnv* env, jobject vthread, jboolean last_unmount))
-  assert(thread->is_in_VTMT(), "VTMT sanity check");
 #if INCLUDE_JVMTI
+  assert(thread->is_in_VTMT(), "VTMT sanity check");
   JvmtiVTMTDisabler::finish_VTMT(vthread, 1);
+#else
+  fatal("Should only be called with JVMTI enabled");
 #endif
 JVM_END
--- a/src/hotspot/share/prims/jvm.cpp
+++ b/src/hotspot/share/prims/jvm.cpp
@@ -3888,11 +3888,14 @@ JVM_ENTRY_NO_ENV(jint, JVM_FindSignal(const char *name))
 JVM_END
 
 JVM_ENTRY(void, JVM_VirtualThreadMountBegin(JNIEnv* env, jobject vthread, jboolean first_mount))
+#if INCLUDE_JVMTI
   JvmtiVTMTDisabler::start_VTMT(vthread, 0);
+#endif
 JVM_END
 
 JVM_ENTRY(void, JVM_VirtualThreadMountEnd(JNIEnv* env, jobject vthread, jboolean first_mount))
   assert(thread->is_in_VTMT(), "VTMT sanity check");
+#if INCLUDE_JVMTI
   JvmtiVTMTDisabler::finish_VTMT(vthread, 0);
   oop vt = JNIHandles::resolve(vthread);
 
@@ -3913,9 +3916,11 @@ JVM_ENTRY(void, JVM_VirtualThreadMountEnd(JNIEnv* env, jobject vthread, jboolean
   if (JvmtiExport::should_post_vthread_mount()) {
     JvmtiExport::post_vthread_mount(vthread);
   }
+#endif
 JVM_END
 
 JVM_ENTRY(void, JVM_VirtualThreadUnmountBegin(JNIEnv* env, jobject vthread, jboolean last_unmount))
+#if INCLUDE_JVMTI
   HandleMark hm(thread);
   Handle ct(thread, thread->threadObj());
 
@@ -3938,9 +3943,12 @@ JVM_ENTRY(void, JVM_VirtualThreadUnmountBegin(JNIEnv* env, jobject vthread, jboo
 
   assert(!thread->is_in_VTMT(), "VTMT sanity check");
   JvmtiVTMTDisabler::start_VTMT(vthread, 1);
+#endif
 JVM_END
 
 JVM_ENTRY(void, JVM_VirtualThreadUnmountEnd(JNIEnv* env, jobject vthread, jboolean last_unmount))
   assert(thread->is_in_VTMT(), "VTMT sanity check");
+#if INCLUDE_JVMTI
   JvmtiVTMTDisabler::finish_VTMT(vthread, 1);
+#endif
 JVM_END
--- a/src/hotspot/share/prims/jvmtiThreadState.cpp
+++ b/src/hotspot/share/prims/jvmtiThreadState.cpp
@@ -82,8 +82,10 @@ JvmtiThreadState::JvmtiThreadState(JavaThread* thread, oop thread_oop)
   _earlyret_oop = NULL;
 
   _jvmti_event_queue = NULL;
+#if INCLUDE_JVMTI
   _is_in_VTMT = false;
   _is_virtual = false;
+#endif
 
   _thread_oop_h = OopHandle(JvmtiExport::jvmti_oop_storage(), thread_oop);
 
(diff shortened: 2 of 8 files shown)


src/hotspot/share/prims/jvm.cpp
```diff
@@ -3913,9 +3916,11 @@ JVM_ENTRY(void, JVM_VirtualThreadMountEnd(JNIEnv* env, jobject vthread, jboolean
   if (JvmtiExport::should_post_vthread_mount()) {
     JvmtiExport::post_vthread_mount(vthread);
   }
+#endif
```
AlanBateman, line 3919: It would be bug if these functions were called when JVMTI isn't compiled in so maybe we should put an #else ShouldNotReachHere() to catch it.

plummercj: I think that makes sense. At first I was thinking why not just #ifdef out the whole API, but these APIs are mapped natives for `VirtualThread.notifyJvmtiMountBegin()` and `VirtualThread.notifyJvmtiMountEnd()`. Calls to them come from java, but should only happen if jvmti is enabled.

src/hotspot/share/prims/jvm.cpp
```diff
@@ -3938,9 +3943,12 @@ JVM_ENTRY(void, JVM_VirtualThreadUnmountBegin(JNIEnv* env, jobject vthread, jboo
...
   assert(!thread->is_in_VTMT(), "VTMT sanity check");
   JvmtiVTMTDisabler::start_VTMT(vthread, 1);
+#endif
 JVM_END
 
 JVM_ENTRY(void, JVM_VirtualThreadUnmountEnd(JNIEnv* env, jobject vthread, jboolean last_unmount))
   assert(thread->is_in_VTMT(), "VTMT sanity check");
+#if INCLUDE_JVMTI
   JvmtiVTMTDisabler::finish_VTMT(vthread, 1);
+#endif
```
plummercj, line 3953: `ShouldNotReachHere()` here also?

src/hotspot/share/prims/jvmtiThreadState.hpp
```diff
@@ -95,8 +95,8 @@ class JvmtiVTMTDisabler {
   void set_self_suspend() { _self_suspend = true; }
   static void start_VTMT(jthread vthread, int callsite_tag);
   static void finish_VTMT(jthread vthread, int callsite_tag);
-  static int  VTMT_disable_count() { return _VTMT_disable_count; };
-  static int  VTMT_count() { return _VTMT_count; }
+  static int  VTMT_disable_count();
+  static int  VTMT_count();
```
plummercj, line 99: I'm not sure what is being fixed here.

src/hotspot/share/prims/jvm.cpp
sspitsyn: I agree, it is a good idea. Of course, it makes sense to add for all 4 native functions:
 JVM_VirtualThreadMountBegin, JVM_VirtualThreadMountEnd,
 JVM_VirtualThreadUnmountBegin, JVM_VirtualThreadUnmountEnd,

sspitsyn (approved):
Aleksey,
I've inlined some comments.
Other than that it looks okay to me.
In fact, it is not easy to identify what else has to be under `#INCLUDE_JVMTI`.
Thanks,
Serguei

src/hotspot/share/runtime/thread.cpp
```diff
@@ -1777,13 +1777,17 @@ void JavaThread::send_thread_stop(oop java_throwable)  {
...
     assert(JvmtiVTMTDisabler::VTMT_disable_count() == 0, "must be 0");
+#endif
   }
 }
 
 void JavaThread::set_is_VTMT_disabler(bool val) {
   _is_VTMT_disabler = val;
+#if INCLUDE_JVMTI
   assert(JvmtiVTMTDisabler::VTMT_count() == 0, "must be 0");
+#endif
```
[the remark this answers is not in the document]
plummercj: I think these APIs should be completely #ifdef'd out, and you also need to wrap the declarations with JVMTI_ONLY macros or append NOT_JVMTI_RETURN to the declarations.

src/hotspot/share/prims/jvmtiThreadState.hpp
shipilev: As stated in PR: "jvmtiExport and jvmtiThreadState headers are usually included even with JVMTI turned off, but .cpp would not be compiled without JVMTI. Therefore, non-trivial implementations should go into .cpp." You cannot put the accessor bodies here, because the `_VTMT_count` and `_VTMT_disable_count` symbols would be inaccessible without `.cpp` compiled, and you'll get linker errors.

src/hotspot/share/prims/jvm.cpp
shipilev: All right, see new commit.

src/hotspot/share/prims/jvmtiThreadState.hpp
plummercj: There are a couple of other choices here. One is like the following example:
```
  inline static bool should_post_class_file_load_hook()           {
    JVMTI_ONLY(return _should_post_class_file_load_hook);
    NOT_JVMTI(return false;)
  }
```
However I don't think anyone should actually be calling any VTMT code when JVMTI is disabled. So you could instead put the entire thing (including declaration) in a `#ifdef INCLUDE_JVMTI`. You would then need to also #ifdef around the callers, or use JVMTI_ONLY. If there are many, then go with my first suggestion.

src/hotspot/share/runtime/thread.hpp
```diff
@@ -1247,8 +1249,23 @@ class JavaThread: public Thread {
...
+  }
+
+  bool is_in_VTMT() const {
+#if INCLUDE_JVMTI
+    return _is_in_VTMT;
+#else
+    fatal("Should only be called with JVMTI enabled");
+    return false;
+#endif
+  }
```
plummercj, line 1268: If you #ifdef out the API, then you know it's not being called, thus no need for #else sections and calls to `fatal()`.

src/hotspot/share/runtime/thread.cpp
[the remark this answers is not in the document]
plummercj: If you #ifdef out the API, then you know it's not being called, thus no need for #else sections and calls to fatal().

src/hotspot/share/prims/jvmtiThreadState.hpp
shipilev: I don't see why we should bother. These methods are only used in asserts, AFAICS, so the performance here is irrelevant. Moving these definitions out of `.hpp` matches the rest of the `JvmtiVTMTDisabler`. There are no other uses of `INCLUDE_JVMTI` in `jvmtiThreadState.hpp`, as it is assumed, I think, that you could include it without JVMTI enabled. If JVMTI is disabled in build time, using these methods would produce linker errors.  Sorry, but I disagree with these suggestions.

src/hotspot/share/runtime/thread.hpp
shipilev: All right, that is indeed cleaner. Please see new commit.

AlanBateman:
The latest comment (9ba55ed) looks fine.

shipilev:
> The latest comment ([9ba55ed](https://github.com/openjdk/loom/commit/9ba55eda41cb018ba34f8a791fa34a4b0517d318)) looks fine.

What are the rules here, should I wait for @plummercj to ack this too?

AlanBateman:
> What are the rules here, should I wait for @plummercj to ack this too?

@sspitsyn and @plummercj had comments so maybe give them a chance to see the updated change. I think what you have is okay, you addressed my comment, and we can always tweak it later. 

Minimal VM was mostly for embedded targets in JDK 8, I don't have a good sense on how many people build and use it these days.

src/hotspot/share/prims/jvmtiThreadState.cpp
```diff
@@ -82,8 +82,10 @@ JvmtiThreadState::JvmtiThreadState(JavaThread* thread, oop thread_oop)
   _earlyret_oop = NULL;
 
   _jvmti_event_queue = NULL;
+#if INCLUDE_JVMTI
   _is_in_VTMT = false;
   _is_virtual = false;
+#endif
```
sspitsyn, line 88: I'm not sure this change is really needed.
My understanding is that the `jvmtiThreadState.cpp` is fully excluded in `non-INCLUDE_JVMTI` mode.

sspitsyn (approved):
I've inlined just one comment.
Other than that the update looks good to me.
Thanks,
Serguei

shipilev:
All right, I would like to integrate this soon. @plummercj had not yet replied for the latest version, but we can amend stuff after integration, if problems arise?

plummercj:
> All right, I would like to integrate this soon. @plummercj had not yet replied for the latest version, but we can amend stuff after integration, if problems arise?

Hmm. I did reply yesterday, but my comment is shown as "pending" and I'm not sure how to resolve that. I'll repost it.

shipilev:
> Hmm. I did reply yesterday, but my comment is shown as "pending" and I'm not sure how to resolve that. I'll repost it.

If you started a review, and then put a comment, it would be "Pending" until you finalize the review with either "Approve", "Change Requested" or "Comment".

shipilev:
GitHub would not let me to reply to the conversation here: https://github.com/openjdk/loom/pull/67#discussion_r718781592 -- so I reply in new comment:

> If they are only used in asserts, then shouldn't they have a #ifdef ASSERTS around them? And there are implicit #ifdef INCLUDE_JVMTI uses in jvmtiThreadState.hpp:

I believe their use in asserts is mostly due to external code accessing the `JvmtiVTMTDisabler` fields. `JvmtiVTMTDisabler` itself accesses the fields directly, because it can, and it does it on non-debug paths. So, while current accesses are from asserts, they do not look like debug-specific accessors. If they are, it would be awkward not to make the fields themselves `DEBUG_ONLY`, but that goes against the `JvmtiVTMTDisabler` internal use.

In other words, I don't see problems with the proposed form. But I see a productivity problem with spending more time polishing this PR. Please accept this form, and let us move on.

plummercj (approved):

shipilev:
Thank you!

/integrate

PR #69, opened 2021-09-27
Fix Optimized builds

shipilev:
The usual trap in Hotspot code is the belief that `ASSERT <=> !PRODUCT`, but that's not actually true. You would not see this in regular release (`!ASSERT`, `PRODUCT`) or fastdebug (`ASSERT`, `!PRODUCT`) builds, but you would see it in optimized (`!ASSERT`, `!PRODUCT`) builds. See for example GHA: https://github.com/shipilev/loom/runs/3717294006?check_suite_focus=true

The use of `ASSERT` and `PRODUCT` should be consistent to avoid build failures in optimized builds.

Additional testing:
 - [x] Linux x86_64 optimized now builds
 - [x] Linux x86_64 release still builds
 - [x] Linux x86_64 fastdebug still builds

---------
### Progress

### Reviewers
 * [Vladimir Kozlov](https://openjdk.java.net/census#kvn) (@vnkozlov - Committer) ⚠️ Review applies to 9070ba6a6f18bffdc1414f319081122f006bbd37

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/69/head:pull/69` \
`$ git checkout pull/69`

`$ git checkout pull/69` \
`$ git pull https://git.openjdk.java.net/loom pull/69/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 69`

`$ git pr show -t 69`

</details>

<a href="https://git.openjdk.java.net/loom/pull/69.diff">https://git.openjdk.java.net/loom/pull/69.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/instanceStackChunkKlass_x86.inline.hpp
+++ b/src/hotspot/cpu/x86/instanceStackChunkKlass_x86.inline.hpp
@@ -170,7 +170,7 @@ class SmallRegisterMap {
 public:
   static constexpr SmallRegisterMap* instance = nullptr;
 private:
-  static void assert_is_rbp(VMReg r) PRODUCT_RETURN 
+  static void assert_is_rbp(VMReg r) NOT_DEBUG({ return; })
                                      DEBUG_ONLY({ assert (r == rbp->as_VMReg() || r == rbp->as_VMReg()->next(), "Reg: %s", r->name()); })
 public:
   // as_RegisterMap is used when we didn't want to templatize and abstract over RegisterMap type to support SmallRegisterMap
--- a/src/hotspot/cpu/x86/instanceStackChunkKlass_x86.inline.hpp
+++ b/src/hotspot/cpu/x86/instanceStackChunkKlass_x86.inline.hpp
@@ -170,7 +170,7 @@ class SmallRegisterMap {
 public:
   static constexpr SmallRegisterMap* instance = nullptr;
 private:
-  static void assert_is_rbp(VMReg r) NOT_DEBUG({ return; })
+  static void assert_is_rbp(VMReg r) NOT_DEBUG_RETURN
                                      DEBUG_ONLY({ assert (r == rbp->as_VMReg() || r == rbp->as_VMReg()->next(), "Reg: %s", r->name()); })
 public:
   // as_RegisterMap is used when we didn't want to templatize and abstract over RegisterMap type to support SmallRegisterMap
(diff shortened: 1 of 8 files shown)


shipilev:
Any takers?

vnkozlov (approved):
Seems fine to me.

shipilev:
Thank you! @AlanBateman -- should I wait for any Loom Reviewers to ack PRs like these?

AlanBateman:
> Thank you! @AlanBateman -- should I wait for any Loom Reviewers to ack PRs like these?

@pron, do you want to look at this?

pron:
Can you explain this change? Why should fields and methods that only appear in assertions be guarded by !PRODUCT rather than ASSERT?

shipilev:
> Can you explain this change? Why should fields and methods that only appear in assertions be guarded by !PRODUCT rather than ASSERT?

AFAICS, everything that is `print_*` can be used from debugging printouts, not only from asserts, so it needs to be `!PRODUCT`. I can tell more if you pick any method as the example.

shipilev:
That said, I can rewrite all `PRODUCT` defines to `ASSERT`/`DEBUG`, if that was the intent. I think the current patch uses `!PRODUCT` and `ASSERT`/`DEBUG` inconsistently, so we can lean into one side more heavily, e.g. drop all `PRODUCT` thingies...

shipilev:
Forgot to say this: notice that `log_develop_trace` is `!PRODUCT` rather than `ASSERT`. So if you reference a field in that logging statement, you need to either make that field `NOT_PRODUCT`, or wrap the `log_develop_trace` into `DEBUG_ONLY`. I believe `NOT_PRODUCT` is less awkward.

shipilev:
See new commit for a bit cleaner patch. It follows the simple guideline: if a field/method is used in any tracing/printout method, it is `!PRODUCT`. The methods that are assert-only are moved to `ASSERT`.

pron:
Ah, OK. I understand the logic now, but it means that the guard will need to be revisited once the traces are cleaned up. You can integrate. Is there a logging mode that is guarded by ASSERT (that would be the most common intent in this code)?

shipilev:
> Ah, OK. I understand the logic now, but it means that the guard will need to be revisited once the traces are cleaned up. 

Yes.

> Is there a logging mode that is guarded by ASSERT (that would be the most common intent in this code)?

I don't think so. IIRC, the purpose of "optimized" was to dump the diagnostic output without suffering the performance loss of fastdebug, so it stands to reason why logging is `!PRODUCT`. But, it is a recurrent theme to ditch the "optimized" builds -- https://bugs.openjdk.java.net/browse/JDK-8183287 -- which should put this problem where it belongs, into the dustbin.

shipilev:
Linux x86_64 `release`, `fastdebug`, `optimized` builds are passing in GHA.

/integrate

PR #72, opened 2021-10-05
Loom: PPC64, S390, ARM32, Zero stubs

shipilev:
This WIP provides enough stubs to make Loom cross-compilable to PPC64, S390, ARM32 and Zero. This should be enough to get the sense where the shared code actually platform-specific (referencing stuff from `frame` that is platform-specific), what should be implemented for a platform, and also should be enough to get GHA build checks green.

Note that ARM32 builds would not be clean until #65 is integrated with 32-bit cleanliness fixes.


pron:
Good to go. Thank you!

PR #81, opened 2021-11-09
Add flag to prevent multiple deopts on the same nmethod.

coleenp:
I added a flag to prevent the same nmethod from being optimized multiple times.  It might be better to have a new MarkForDeoptimizationStatus enum but wasn't sure how that would interact with the deoptimize_update state.  Advice welcome.
I also don't know why the sweeper is so slow to get rid of these entries.
This makes ctw_2 test run in an hour on my machine, just like mainline jdk/jdk (machine is slow!) in fastdebug.

---------
### Progress

### Reviewers
 * [Ron Pressler](https://openjdk.java.net/census#rpressler) (@pron - **Reviewer**) ⚠️ Review applies to cb8b8925a13aa0743a3ed7f54fc73b9eabc9576a
 * [Rickard Bäckman](https://openjdk.java.net/census#rbackman) (@rickard - Committer) ⚠️ Review applies to 0d998c613b89d5af477f4e0b1fe3eefc2482f588

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/81/head:pull/81` \
`$ git checkout pull/81`

`$ git checkout pull/81` \
`$ git pull https://git.openjdk.java.net/loom pull/81/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 81`

`$ git pr show -t 81`

</details>

<a href="https://git.openjdk.java.net/loom/pull/81.diff">https://git.openjdk.java.net/loom/pull/81.diff</a>

</details>


pron (approved):
Seems fine to me, although I agree that adding another value to `MarkForDeoptimizationStatus` would look cleaner and make more sense if it were possible. I don't know the answer to that.

coleenp:
I had a look again and can add the flag to MarkForDeoptimizationStatus which I left as a u1.

pron (approved):

rickard (approved):
Looks good.

coleenp:
Erik has a different change that's going into mainline first.

coleenp:
Reopening.  This might be the simplest fix for the multiple deopt problem for loom running make_deoptimized() on nmethods more than once.  Alternative implementations need architectural changes to code cache walking and my other change depends on not needing the CompiledIC_lock for nop patching in the nmethods, which is semi-safe but may or may not interact with the oopmap patching for these nops.  ie. that needs more investigation but fix number 2 is a big less provably correct.

coleenp:
ctw is still really slow. Takes about 70 minutes on loom and 10 on mainline.

coleenp:
I added some logging to CodeCache::make_marked_nmethods_not_entrant().
For mainline, it's called 395581 times -> max 2946 not entrant found -> 4072 max nmethods iterated
For loom,       it's called 337990 times -> max 1367 not entrant found -> 94950 max nmethods iterated

1367 is with this change.  The 94K nmethods iterated seems to be why CTW takes 7x longer for loom than mainline.

/integrate

PR #83, opened 2021-12-20
Update running count of ForkJoinPool when thread pinned

miao-zheng:
In the test case of this PR, when vt0 invoke System.out.println(), it will invoke managedBlock eventually, and create a new ForkJoinWorkerThread. The call stack is :
```
at java.base/java.util.concurrent.ForkJoinPool.tryCompensate(ForkJoinPool.java:1789)
        at java.base/java.util.concurrent.ForkJoinPool.compensatedBlock(ForkJoinPool.java:3448)
        at java.base/jdk.internal.misc.Blocker.managedBlock(Blocker.java:185)
        at java.base/java.io.FileOutputStream.write(FileOutputStream.java:365)
        at java.base/java.io.BufferedOutputStream.flushBuffer(BufferedOutputStream.java:126)
        at java.base/java.io.BufferedOutputStream.lockedFlush(BufferedOutputStream.java:251)
        at java.base/java.io.BufferedOutputStream.flush(BufferedOutputStream.java:239)
        at java.base/java.io.PrintStream.lockedWrite(PrintStream.java:655)
        at java.base/java.io.PrintStream.write(PrintStream.java:633)
        at java.base/sun.nio.cs.StreamEncoder.writeBytes(StreamEncoder.java:308)
        at java.base/sun.nio.cs.StreamEncoder.implFlushBuffer(StreamEncoder.java:404)
        at java.base/sun.nio.cs.StreamEncoder.lockedFlushBuffer(StreamEncoder.java:126)
        at java.base/sun.nio.cs.StreamEncoder.flushBuffer(StreamEncoder.java:113)
        at java.base/java.io.OutputStreamWriter.flushBuffer(OutputStreamWriter.java:176)
        at java.base/java.io.PrintStream.lockedWriteln(PrintStream.java:859)
        at java.base/java.io.PrintStream.writeln(PrintStream.java:836)
        at java.base/java.io.PrintStream.println(PrintStream.java:1178)
        at VTCompensate$1.run(VTCompensate.java:45)
        at java.base/java.lang.VirtualThread.run(VirtualThread.java:270)
        at java.base/java.lang.VirtualThread$VThreadContinuation.lambda$new$0(VirtualThread.java:155)
        at java.base/jdk.internal.vm.Continuation.enter0(Continuation.java:377)
        at java.base/jdk.internal.vm.Continuation.enter(Continuation.java:370)
```

After create a new ForkJoinWorkerThread, we have two ForkJoinWorkerThread(use -Djdk.defaultScheduler.parallelism=1 for initialization), and vt1 pinned a ForkJoinWorkerThread.

When main thread unlock Reentrantlock, vt0 will re-submit to ForkJoinPool and we have an extra working thread to run vt0. 

But we find the extra compensate working thread not work, it is parked. The call stack is :
```
"ForkJoinPool-1-worker-2" #18 [31814] daemon prio=5 os_prio=0 cpu=0.42ms elapsed=119.94s tid=0x00007f8cf8005050 nid=31814 waiting on condition  [0x00007f8d4012c000]
   java.lang.Thread.State: WAITING (parking)
        at jdk.internal.misc.Unsafe.park(java.base@19-internal/Native Method)
        at java.util.concurrent.locks.LockSupport.park(java.base@19-internal/LockSupport.java:371)
        at java.util.concurrent.ForkJoinPool.awaitWork(java.base@19-internal/ForkJoinPool.java:1706)
        at java.util.concurrent.ForkJoinPool.runWorker(java.base@19-internal/ForkJoinPool.java:1603)
        at java.util.concurrent.ForkJoinWorkerThread.run(java.base@19-internal/ForkJoinWorkerThread.java:165)
```

The problem is when re-submit vt-0 to ForkJoinPool, it will call signalWork() of ForkJoinPool and signalWork() do nothing because current c > 0L and the judgement of c < 0L will jump over the loop(c < 0L means RC < parallelism, in our test case, parallelism is 1 and current running count is 1)

Actually, current running count should be 0 now because pinned thread can not running other task, so we think should update running count of ForkJoinPool before and after pin.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/83/head:pull/83` \
`$ git checkout pull/83`

`$ git checkout pull/83` \
`$ git pull https://git.openjdk.java.net/loom pull/83/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 83`

`$ git pr show -t 83`

</details>

<a href="https://git.openjdk.java.net/loom/pull/83.diff">https://git.openjdk.java.net/loom/pull/83.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/prims/jvmtiExport.cpp
+++ b/src/hotspot/share/prims/jvmtiExport.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2003, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2003, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -1266,14 +1266,13 @@ void JvmtiExport::check_suspend_at_safepoint(JavaThread *thread) {
   oop vt = thread->jvmti_vthread();
 
   if (vt != NULL && java_lang_VirtualThread::is_instance(vt)) {
-    HandleMark hm(thread);
-    Handle vth = Handle(thread, vt);
+    int64_t id = java_lang_Thread::thread_id(vt);
 
     ThreadBlockInVM tbivm(thread);
     MonitorLocker ml(JvmtiVTMT_lock, Mutex::_no_safepoint_check_flag);
 
     // block while vthread is externally suspended
-    while (JvmtiVTSuspender::is_vthread_suspended(vth())) {
+    while (JvmtiVTSuspender::is_vthread_suspended(id)) {
       ml.wait();
     }
   }


AlanBateman:
I don't think we should do this, at least not as proposed. We either consistently compensate or not compensate, it's just too unpredictable to try to make use of idle workers when parallelism has not been increased.  In PR 80, Ron argued that pinning should not compensate so we either have to re-visit that issue or do nothing. 

I also have a concern with the APIs exposed because (a) the FJP API doesn't expose anything about release counts, and (b) the Thread object for carrier threads is not exposed.

miao-zheng:
Thanks for your response.

I found there are many logic of managedBlock added to Loom. And there are many original paths to managedBlock which will invoke tryCompensate eventually.

The problem of this PR is compensated threads may not be used when RC of FJP greater than -Djdk.defaultScheduler.parallelism while pin happens.

AlanBateman:
> The problem of this PR is compensated threads may not be used when RC of FJP greater than -Djdk.defaultScheduler.parallelism while pin happens.

The behaviour you observe is expected. You've set parallelism to 1. It increases temporarily because one of the virtual threads invokes code that does a managed block. It then decreases back to 1 when that is done. I think what you are asking is to re-visit the question as to whether pinning should increase parallelism. Ron argued in the other PR that it should not on the grounds that if you are dependent on that then the system will eventually wedge anyway. For cases like this it would be better to change the code to eliminate the pinning, we hope this is temporary until Java monitors are-implemented.

miao-zheng:
Thanks for your response.

Eliminating pin is absolutely better. Except supporting Java monitors, I still have a question that how to support native frame or what the plan is with native frame?

I have another question that we should think a bit of pin is tolerable or not?If a bit of pin is tolerable, then maybe we should make pin work better with scheduler.

PR #91, opened 2022-02-28
8279653: compiler/codecache/stress/OverloadCompileQueueTest.java failing in loom repo

coleenp:
I fixed the test to disable compiling Thread.sleep() and all the methods it seems to call, and this makes the test pass.  I don't know if there's a better way to disable compilation like this that would be less fragile (ie if Thread.sleep() calls something new in the future).  See bug comments for details.  @rickard  do you know?  Thanks!


coleenp:
Thanks @AlanBateman for pointing to the difference in Thread.sleep() and suggestion.

rickard (approved):

coleenp:
Thanks Rickard.
(I wonder if I hit the "Merge pull request" button, or integrate)
/integrate

AlanBateman (approved):

PR #97, opened 2022-03-07
Better jdb support for vthreads. Improved jdb vthread testing support.

plummercj:
This PR contains a set of some what disparate but related set of changes to JDB, JDI, and nsk jdb tests and test support (Jdb.java).

jdb: Made the “kill” command support `ThreadReference.stop()` throwing `IllegalThreadStateException` for vthreads, because this is how `ThreadReference.stop()` is now spec’d. This change is in `Commands.java`. The rest of the changes are just cleanup. One of the changes in the `kill001` test is to test for this (see below)

jdb: Use `PlatformThreadOnlyFilters` to prevent `ThreadStart/Death` event for vthreads. Instead detect vthreads when events are received on them. The goal here is to prevent jdb from getting a flood of `ThreadStart/Death` events, and getting bogged down with them. With the `PlatformThreadOnlyFilters` in place, this means jdb is not notified when a vthread dies, even one that it learned about later when an event came in on it. For these each of these “learned” vthreads a separate `ThreadDeathRequest` is needed that uses the thread itself as a `ThreadFilter`. This way jdb can tell when these vthreads die. This part is handled in `EventHandler.java`. `VMConnection.java` is where `PlatformThreadOnlyFilters` are added.

jdb: Add `-trackallvthreads` option. This prevents the use of the `PlatformThreadsOnlyFilters` described above, and instead relies on jdb getting the initial set of vthreads from the debug agent via `VM.allThreads()` (by turning on the `enumeratevthreads` option), and by getting a `ThreadStart/Death` event for every vthread created/destroyed after connecting. This is basially how jdb used to work by default before any changes in this PR. There are 3 tests that rely on this behaviour and need this flag. They are `kill001`, `threads002`, and `trace001` (see below). Many more would rely on it if they were converted to use vthreads as these tests were. The bulk of the changes for this support comes from having to pass `trackAllVthreads` through all sorts of constuctors/inits. I wish there were a better way, but I couldn’t find a place to store this value as a static so it is easy to access. If you see a better way, let me know.

jdb: Track vthreads separately. jdb keeps track of every thread by creating a `ThreadInfo` for each and storing it in a `List<ThreadInfo>` called `threads`. I decided to store vthreads in a separate list. The reason is the list is needed by the `threads` command. It doesn’t use the `threads` list, but instead walks the `ThreadGroup` hierarchy and prints out the threads within each `ThreadGroup`. This doesn’t capture vthreads since they are not returned by `ThreadGroupReference.threads()`, even for the `VirtualThreads` group. So after the `threads` command has done its usual `ThreadGroup` traversal, it now also dumps all the vthreads. This is pretty much all hidden in `ThreadIterator.next()`, which automically transitions to getting threads from the vthread list after traversing all the `ThreadGroups`.

`nsk/share/jdb/Jdb.java`: Added support for `getThreadIdsByName()`: There are many jdb tests that use `Jdb.getThreadIds(<classname>)` to get all threads in the `threads` output that are an instance of `<classname>`. The problem is that after converting a test to support a vthread mode for all debuggee threads, the name of the class in the output will always be `java.lang.VirtualThread` (or `java.lang.Thread` when not in vthread mode), and not the test specific class name, which is usually something like `MyThread`.  Making `getThreadIds()` search for `java.lang.VirtualThreads` usually ends up including threads not wanted (ones that are not `MyThread`) such as the main test thread. So `getThreadIdsByName()` was added to allow searching by the thread name (actually just the prefix of the name) rather than its class. As an example, the `threads` command output used to contain a line such as:

`   (nsk.jdb.kill.kill001.MyThread)693            Thread-0            waiting in a monitor`

Now it contains:

`   (java.lang.Thread)693                         MyThread-0          waiting in a monitor`

So instead of searching for “(MyThread)” and grabbing the threadId that comes right after it, `getThreadIdsByName()` first finds all the lines of output that contain “nsk.jdb.kill.kill001.MyThread”, and then for those lines only gets the threadID that is right after “(java.lang.Thread)” or “(java.lang.VirtualThread)”.

jdb `kill001` test: Convert to support all debuggee threads being vthreads by using `JDIThreadFactory`. Enable new jdb `-trackallvthreads` option. Use new `getThreadIdsByName()` support instead of `getThreadIds()`. Allow/expect the `kill` command to fail with "Illegal thread state” when in vthread mode.

jdb `threads002` test: Convert to support all debuggee threads being vthreads by using `JDIThreadFactory`. Enable new jdb `-trackallvthreads` option. I forget why I decided to convert this test. Seems it was useful for testing something I was working on for this PR, but I can’t recall. Maybe it was so I could clone it for the new `threads003` test below.

jdb `trace001` test: Convert to support all debuggee threads being vthreads by using `JDIThreadFactory`. Enable new jdb `-trackallvthreads` option. This is a test I had converted long ago when I first experimented with the `getThreadIdsByName()` support, so I included it with this PR.

jdb `threads003` test: This is a new thread test that tests jdb’s ability to learn about a thread when an event arrives on it. It creates 5 threads, two of which receive a breakpoint. It looks at the `threads` output and checks that only those two vthreads are present. I also used this test to help confirm that the `ThreadDeath` events are only received for these two vthreads. This was done manually by looking for the “TDR removed” output generated by `EventHandler.threadDeathEvent()`. I didn’t put this into the test itself since that println is just for debugging and is now commented out.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/97/head:pull/97` \
`$ git checkout pull/97`

`$ git checkout pull/97` \
`$ git pull https://git.openjdk.java.net/loom pull/97/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 97`

`$ git pr show -t 97`

</details>

<a href="https://git.openjdk.java.net/loom/pull/97.diff">https://git.openjdk.java.net/loom/pull/97.diff</a>

</details>

### DIFF
--- a/src/jdk.jdi/share/classes/com/sun/tools/example/debug/tty/EventHandler.java
+++ b/src/jdk.jdi/share/classes/com/sun/tools/example/debug/tty/EventHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2016, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -37,6 +37,8 @@ package com.sun.tools.example.debug.tty;
 import com.sun.jdi.*;
 import com.sun.jdi.event.*;
 import com.sun.jdi.request.EventRequest;
+import com.sun.jdi.request.EventRequestManager;
+import com.sun.jdi.request.ThreadDeathRequest;
 
 public class EventHandler implements Runnable {
 
@@ -46,10 +48,12 @@ public class EventHandler implements Runnable {
     boolean completed = false;
     String shutdownMessageKey;
     boolean stopOnVMStart;
+    boolean trackAllVthreads;
 
-    EventHandler(EventNotifier notifier, boolean stopOnVMStart) {
+    EventHandler(EventNotifier notifier, boolean stopOnVMStart, boolean trackAllVthreads) {
         this.notifier = notifier;
         this.stopOnVMStart = stopOnVMStart;
+        this.trackAllVthreads = trackAllVthreads;
         this.thread = new Thread(this, "event-handler");
         this.thread.start();
     }
@@ -95,6 +99,32 @@ public class EventHandler implements Runnable {
 
     private boolean handleEvent(Event event) {
         notifier.receivedEvent(event);
+
+        /*
+         * See if the event thread is a vthread that we need to start tracking.
+         */
+        ThreadReference eventThread = null;
+        if (event instanceof ClassPrepareEvent) {
+            eventThread = ((ClassPrepareEvent)event).thread();
+        } else if (event instanceof LocatableEvent) {
+            eventThread = ((LocatableEvent)event).thread();
+        }
+        if (eventThread != null) {
+            // This is a vthread we haven't seen before, so add it to our list.
+            assert eventThread.isVirtual();
+            boolean added = ThreadInfo.addThread(eventThread);
+            // If we added it, we need to make sure it eventually gets removed. Usually
+            // this happens when the ThreadDeathEvent comes in, but if !trackAllVthreads,
+            // then the ThreadDeathReqest was setup to filter out vthreads. So we'll need
+            // to create a special ThreadDeathReqest just to detect when this vthread dies.
+            if (added && !trackAllVthreads) {
+                EventRequestManager erm = Env.vm().eventRequestManager();
+                ThreadDeathRequest tdr = erm.createThreadDeathRequest();
+                tdr.addThreadFilter(eventThread);
+                tdr.enable();
+                //System.out.println("TDR added: " + thread);
+            }
+        }
 
         if (event instanceof ExceptionEvent) {
             return exceptionEvent(event);
@@ -264,8 +294,17 @@ public class EventHandler implements Runnable {
     }
 
     private boolean threadDeathEvent(Event event) {
-        ThreadDeathEvent tee = (ThreadDeathEvent)event;
-        ThreadInfo.removeThread(tee.thread());
+        ThreadDeathEvent tde = (ThreadDeathEvent)event;
+        ThreadReference thread = tde.thread();
+        ThreadInfo.removeThread(thread);
+
+        if (!trackAllVthreads && thread.isVirtual()) {
+            // Remove the ThreadDeathRequest used for this event since it was created
+            // just to remove this one vthread.
+            EventRequestManager erm = Env.vm().eventRequestManager();
+            erm.deleteEventRequest(event.request());
+            //System.out.println("TDR removed: " + thread);
+        }
         return false;
     }
 
--- a/src/jdk.jdi/share/classes/com/sun/tools/example/debug/tty/EventHandler.java
+++ b/src/jdk.jdi/share/classes/com/sun/tools/example/debug/tty/EventHandler.java
@@ -48,12 +48,12 @@ public class EventHandler implements Runnable {
     boolean completed = false;
     String shutdownMessageKey;
     boolean stopOnVMStart;
-    boolean trackAllVthreads;
+    boolean trackVthreads;
 
-    EventHandler(EventNotifier notifier, boolean stopOnVMStart, boolean trackAllVthreads) {
+    EventHandler(EventNotifier notifier, boolean stopOnVMStart, boolean trackVthreads) {
         this.notifier = notifier;
         this.stopOnVMStart = stopOnVMStart;
-        this.trackAllVthreads = trackAllVthreads;
+        this.trackVthreads = trackVthreads;
         this.thread = new Thread(this, "event-handler");
         this.thread.start();
     }
@@ -118,10 +118,10 @@ public class EventHandler implements Runnable {
                 assert eventThread.isVirtual();
             }
             // If we added it, we need to make sure it eventually gets removed. Usually
-            // this happens when the ThreadDeathEvent comes in, but if !trackAllVthreads,
+            // this happens when the ThreadDeathEvent comes in, but if !trackVthreads,
             // then the ThreadDeathReqest was setup to filter out vthreads. So we'll need
             // to create a special ThreadDeathReqest just to detect when this vthread dies.
-            if (added && !trackAllVthreads) {
+            if (added && !trackVthreads) {
                 EventRequestManager erm = Env.vm().eventRequestManager();
                 ThreadDeathRequest tdr = erm.createThreadDeathRequest();
                 tdr.addThreadFilter(eventThread);
@@ -301,7 +301,7 @@ public class EventHandler implements Runnable {
         ThreadReference thread = tde.thread();
         ThreadInfo.removeThread(thread);
 
-        if (!trackAllVthreads && thread.isVirtual()) {
+        if (!trackVthreads && thread.isVirtual()) {
             // Remove the ThreadDeathRequest used for this event since it was created
             // just to remove this one vthread.
             EventRequestManager erm = Env.vm().eventRequestManager();
--- a/src/jdk.jdi/share/classes/com/sun/tools/example/debug/tty/EventHandler.java
+++ b/src/jdk.jdi/share/classes/com/sun/tools/example/debug/tty/EventHandler.java
@@ -111,8 +111,10 @@ public class EventHandler implements Runnable {
         }
         if (eventThread != null) {
             // This is a vthread we haven't seen before, so add it to our list.
-            assert eventThread.isVirtual();
             boolean added = ThreadInfo.addThread(eventThread);
+            if (added) {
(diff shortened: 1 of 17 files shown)


src/jdk.jdi/share/classes/com/sun/tools/example/debug/tty/EventHandler.java
```diff
@@ -264,8 +294,17 @@ private boolean exceptionEvent(Event event) {
...
+        ThreadDeathEvent tde = (ThreadDeathEvent)event;
+        ThreadReference thread = tde.thread();
+        ThreadInfo.removeThread(thread);
+
+        if (!trackAllVthreads && thread.isVirtual()) {
+            // Remove the ThreadDeathRequest used for this event since it was created
+            // just to remove this one vthread.
+            EventRequestManager erm = Env.vm().eventRequestManager();
+            erm.deleteEventRequest(event.request());
+            //System.out.println("TDR removed: " + thread);
```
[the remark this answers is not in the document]
plummercj: Yes, I plan on removing these. They were useful for making sure it was working properly.

src/jdk.jdi/share/classes/com/sun/tools/example/debug/tty/TTY.java
```diff
@@ -74,6 +74,8 @@
 
     private volatile boolean shuttingDown = false;
 
+    private static boolean trackAllVthreads = false;
```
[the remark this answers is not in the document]
plummercj: Ok.

src/jdk.jdi/share/classes/com/sun/tools/example/debug/tty/TTY.java
```diff
@@ -972,6 +974,8 @@ public static void main(String argv[]) throws MissingResourceException {
                         return;
                     }
                 }
+            } else if (token.equals("-trackallvthreads")) {
```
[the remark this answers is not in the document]
plummercj: I think you  meant `-trackallvthreads`. Do we have a CSR in place yet that I can add too?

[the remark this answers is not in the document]
AlanBateman: >  Do we have a CSR in place yet that I can add too?

Not for the loom repo but we will include need to remember to include it in the eventual CSR when we bring the feature into main line.

src/jdk.jdi/share/classes/com/sun/tools/example/debug/tty/EventHandler.java
```diff
@@ -96,6 +100,32 @@ public void run() {
...
+         */
+        ThreadReference eventThread = null;
+        if (event instanceof ClassPrepareEvent) {
+            eventThread = ((ClassPrepareEvent)event).thread();
+        } else if (event instanceof LocatableEvent) {
+            eventThread = ((LocatableEvent)event).thread();
+        }
+        if (eventThread != null) {
+            // This is a vthread we haven't seen before, so add it to our list.
+            assert eventThread.isVirtual();
```
sspitsyn, line 114: It is not clear why a virtual thread is expected here.
At least, some comment would be nice to have.

sspitsyn: I looked at the jdb changes. They look good to me. I've posed one comment though.

plummercj: Yeah, that got misplaced, probably when I shuffled around some code. I think it would be correct if it was moved below the `addThread()` call and was changed to:

`            assert !(added && !eventThread.isVirtual());`

which is probably more easily read as:

`            assert !added || eventThread.isVirtual());`

First I need to see if I can even get asserts enabled since that one should have been triggered.

plummercj: Actually that's not quite right either. It looks like it needs to be:

```
            if (added) {
                assert eventThread.isVirtual();
            }
```
If this thread was added, it must be an a virtual thread, because if it was not it would have been previously added by a ThreadStart event.

sspitsyn: It is better to add a comment explaining this.
Otherwise, it is hard to understand why this conclusion was made.
Also, why do you think this is not equal of your conditional assert above?:
   assert !added || eventThread.isVirtual());

plummercj: > It is better to add a comment explaining this.
> Otherwise, it is hard to understand why this conclusion was made.

I added some comments.

> Also, why do you think this is not equal to your conditional assert above?:
> assert !added || eventThread.isVirtual());

Actually that does work, but I think it is easier to read with the "if" statement.

test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
```diff
@@ -50,12 +50,13 @@
  * @run main/othervm
  *      nsk.jdb.kill.kill001.kill001
  *      -arch=${os.family}-${os.simpleArch}
- *      -waittime=5
+ *      -waittime=1 -verbose
```
alexmenkov, line 53: please add -verbose at separate line.
And I suppose you changed waittime temporary and forgot to revert it.
The same for other tests

test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
```diff
@@ -98,25 +99,38 @@ protected void runCases() {
...
         }
 
-        for (int i = 0; i <= numThreads; i++) {
+        // Continue the main thread, which is still at the breakpoint. This will resume
+        // all the debuggee threads, allowing the kill to take place.
+        reply = jdb.receiveReplyFor(JdbCommand.cont);
+
+        // Continue each of the threads that received the "kill" exception. Not needed
+        // for the vthread case since they are not actually killed.
+        if (!vthreadMode) for (int i = 0; i < numThreads; i++) {
```
alexmenkov, line 133: if and for at the same line don't look good

src/jdk.jdi/share/classes/com/sun/tools/example/debug/tty/EventHandler.java
```diff
@@ -96,6 +100,35 @@ public void run() {
     private boolean handleEvent(Event event) {
         notifier.receivedEvent(event);
 
+        /*
+         * See if the event thread is a vthread that we need to start tracking.
+         */
+        ThreadReference eventThread = null;
+        if (event instanceof ClassPrepareEvent) {
+            eventThread = ((ClassPrepareEvent)event).thread();
```
alexmenkov, line 108:
```suggestion
        if (event instanceof ClassPrepareEvent evt) {
            eventThread = evt.thread();
```
The same for lines 109-110

test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
plummercj: Ok, will move to separate line. For `-waittime`, 5 seems overkill for all these tests. If anything ever goes wrong with the test, it seems you always end up waiting for a timeout. So first thing I do is change `-waittime` to 1 so I don't have to wait as long when debugging the test. I'd change it to .5 if I could, but it doesn't accept decimals. I could change it back to 5, but I don't see why it needs to be so big. Maybe @lmesnik can comment on this.

src/jdk.jdi/share/classes/com/sun/tools/example/debug/tty/EventHandler.java
plummercj: op

test/hotspot/jtreg/vmTestbase/nsk/jdb/kill/kill001/kill001.java
plummercj: ok

test/hotspot/jtreg/vmTestbase/nsk/jdb/threads/threads002/threads002a.java
```diff
@@ -51,7 +53,8 @@ public int runIt(String args[], PrintStream out) {
         try {
             lock.setLock();
             for (int i = 0; i < numThreads ; i++) {
-                holder[i] = new MyThread(lock);
+                String name = THREAD_NAME + "-" + i;
+                holder[i] = JDIThreadFactory.newThread(new MyThread(lock), name);
```
lmesnik, line 57: Please change MyThread to implements Runnable from extends Thread.

test/hotspot/jtreg/vmTestbase/nsk/jdb/threads/threads003/threads003a.java
```diff
@@ -0,0 +1,131 @@
...
+
+    synchronized void releaseLock() {
+        if (lockSet == true) {
+            lockSet = false;
+            notify();
+        }
+    }
+}
+
+class MyThread extends Thread {
```
lmesnik, line 106: Please change MyThread to implements Runnable from extends Thread.

PR #114, opened 2022-03-22
Small fixes - shipment one

robehn:
Reducing the number of variables, the number of assignment to each variable, white-spaces, renamed some variables, removed unnecessary template, etc...
Except the removal of template 'bottom', the code is the same. (assert is moved up one level)

Passes 1-3 (no new failures)
It seem to be slightly faster than before, but at least on par.

This is not complete at all, but to avoid merge conflict I want to ship, and also make someone happy about removing a template.

---------
### Progress

### Reviewers
 * [Erik Österlund](https://openjdk.java.net/census#eosterlund) (@fisk - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/114/head:pull/114` \
`$ git checkout pull/114`

`$ git checkout pull/114` \
`$ git pull https://git.openjdk.java.net/loom pull/114/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 114`

`$ git pr show -t 114`

</details>

<a href="https://git.openjdk.java.net/loom/pull/114.diff">https://git.openjdk.java.net/loom/pull/114.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/runtime/continuation.cpp
+++ b/src/hotspot/share/runtime/continuation.cpp
@@ -1031,7 +1031,6 @@ private:
   const bool _preempt; // used only on the slow path
 
   intptr_t *_bottom_address;
-  intptr_t *_top_address;
 
   int _size; // total size of all frames plus metadata in words.
   int _align_size;
@@ -1328,7 +1327,6 @@ public:
       f.print_on(&ls);
     }
 
-    _top_address = f.sp();
     frame caller;
     freeze_result res = freeze(f, caller, 0, false, true);
 
--- a/src/hotspot/share/runtime/continuation.cpp
+++ b/src/hotspot/share/runtime/continuation.cpp
@@ -1048,7 +1048,8 @@ public:
   Freeze(JavaThread* thread, ContMirror& mirror, bool preempt) :
       _thread(thread), _cont(mirror), _barriers(false), _preempt(preempt) {
 
-    assert(thread->last_continuation()->entry_sp() == _cont.entrySP(), "");
+    assert(_thread != nullptr, "");
+    assert(_thread->last_continuation()->entry_sp() == _cont.entrySP(), "");
 
     int argsize = _cont.argsize();
     _bottom_address = _cont.entrySP() - argsize;
@@ -1093,10 +1094,9 @@ public:
 
   // Called _after_ the last possible sfepoint during the freeze operation (chunk allocation)
   void unwind_frames() {
-    JavaThread* thread = _thread;
     ContinuationEntry* entry = _cont.entry();
-    ContinuationHelper::maybe_flush_stack_processing(thread, entry);
-    ContinuationHelper::set_anchor_to_entry(thread, entry);
+    ContinuationHelper::maybe_flush_stack_processing(_thread, entry);
+    ContinuationHelper::set_anchor_to_entry(_thread, entry);
   }
 
   template <bool chunk_available>
@@ -1104,7 +1104,7 @@ public:
     if (freeze_fast<chunk_available>(sp)) {
       return freeze_ok;
     }
-    if (_thread != nullptr && _thread->has_pending_exception()) {
+    if (_thread->has_pending_exception()) {
       return freeze_exception;
     }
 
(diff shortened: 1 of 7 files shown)


src/hotspot/share/runtime/continuation.cpp
```diff
@@ -174,10 +174,10 @@ Address |   |                            |    |   Caller is still in the chunk.
...
 extern "C" bool dbg_is_safe(const void* p, intptr_t errvalue); // address p is readable and *(intptr_t*)p != errvalue
 
-template<int x> NOINLINE static bool verify_continuation(oop cont) { return Continuation::debug_verify_continuation(cont); }
-#define VERIFY_CONTINUATION(cont) verify_continuation<__LINE__>((cont))
-template<int x> NOINLINE static bool verify_stack_chunk(oop chunk) { return InstanceStackChunkKlass::verify(chunk); }
-#define VERIFY_STACK_CHUNK(chunk) verify_stack_chunk<__LINE__>((chunk))
+NOINLINE static bool verify_continuation(oop cont) { return Continuation::debug_verify_continuation(cont); }
+#define VERIFY_CONTINUATION(cont) verify_continuation((cont))
+NOINLINE static bool verify_stack_chunk(oop chunk) { return InstanceStackChunkKlass::verify(chunk); }
+#define VERIFY_STACK_CHUNK(chunk) verify_stack_chunk((chunk))
```
coleenp, line 180: Thanks for fixing this.  I know Ron likes to see the asserts have a line number of the caller but there's a PR to fix this and almost all of use some debugger to see where the assert comes from in the call stack, if it's not already in the hs_err file.

src/hotspot/share/runtime/continuation.cpp
```diff
@@ -1130,106 +1134,96 @@ class Freeze : public FreezeBase {
...
   template <bool chunk_available>
   bool freeze_fast(intptr_t* top_sp) {
-    assert(_thread != nullptr, "");
     assert(_cont.chunk_invariant(tty), "");
     assert(!Interpreter::contains(_cont.entryPC()), "");
     assert(StubRoutines::cont_doYield_stub()->frame_size() == ContinuationHelper::frame_metadata, "");
 
     // properties of the continuation on the stack; all sizes are in words
-    intptr_t* const stack_top     = top_sp + ContinuationHelper::frame_metadata;
-    const int       stack_argsize = _cont.argsize();
```
coleenp, line 1160: I like this spelled out in the places it's used.  I wonder if _cont.argsize() is the stacksize of the stack chunk oops.  There are no comments really what it is.
cont_size seems to be the stack size of the entire continuation, ie the sum of all the stack chunk oops?  Dunno.  We're going to need some comments.

src/hotspot/share/runtime/continuation.cpp
```diff
@@ -1130,106 +1134,96 @@ class Freeze : public FreezeBase {
...
       assert(_thread->thread_state() == _thread_in_vm, "");
       assert(!is_chunk_available(top_sp), "");
       assert(_thread->cont_fastpath(), "");
 
-      chunk = allocate_chunk(size + ContinuationHelper::frame_metadata);
-      if (UNLIKELY(chunk == nullptr)) return false; // OOME
-      if (UNLIKELY(!_thread->cont_fastpath()
-                  || _barriers)) { // probably humongous
+      chunk = allocate_chunk(cont_size + ContinuationHelper::frame_metadata);
+      if (UNLIKELY(chunk == nullptr || !_thread->cont_fastpath() || _barriers)) { // OOME/probably humongous
```
coleenp, line 1213: This is just a strange set of conditions.  Why would you allocate_chunk if  _thread->cont_fastpath() is false? or _barriers, whatever that means.  Really don't expect a function like allocate_chunk to have side effects.
Maybe in the latter two conditions, you still want to zero out the chunk in the tail and don't want to remove this.

coleenp:
| and also make someone happy about removing a template.
me!

src/hotspot/share/runtime/continuation.cpp
```diff
@@ -1847,12 +1844,7 @@ class Freeze : public FreezeBase {
     assert(chunk->size() >= stack_size, "chunk->size(): %zu size: %zu", chunk->size(), stack_size);
     assert((intptr_t)chunk->start_address() % 8 == 0, "");
 
-    // TODO PERF: maybe just memset 0, and only set non-zero fields.
-    chunk->clear_flags();
-    chunk->set_gc_mode(false);
-    chunk->set_max_size(0);
-    // chunk->set_pc(nullptr);
-    // chunk->set_argsize(0);
+    init_chunk(chunk);
```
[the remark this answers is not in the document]
robehn: Yes, the allocation returns memory without zeroing it.

src/hotspot/share/runtime/continuation.cpp
robehn: The variable _barriers is set by allocate_chunk, so it must be check afterwards.
If allocation safepoints and there is JVM TI qeury it can 'remove' us as available for fastpath.
If allocation fails due to OOME.

The init_chunk() should always zero enought.

src/hotspot/share/runtime/continuation.cpp
robehn: _cont argsize is normally: inline int StackChunkFrameStream<frame_kind>::stack_argsize()
But it depends what we mean, a single intepreted frame must have the args from caller to be mounted.
But if caller is already mounted we already have those arguments.

src/hotspot/share/runtime/continuation.cpp
coleenp: Ok, thanks for confirming.

src/hotspot/share/runtime/continuation.cpp
coleenp: I guess if the callers are different, you can tell by the caller which is why this is seldom something I've noticed or thought needed fixing.

PR #117, opened 2022-03-23
JcstressRunner: remove artifact version duplication

lepestock:
Tiny style-related fix: allows for modifying the artifact version in only one line.

Testing: some Jcstress tests using that Runner were run locally.


AlanBateman:
Can we close this?

PR #122, opened 2022-03-24
Lower case verify calls

coleenp:
This is another cosmetic change to fix the verify functions so that there is one less null string assert.

This change also removes the copy_alignment template parameter as Patricio, Robbin and I noticed it's unused.  Which is good because then we'd have to wonder when and why it would be less than the default. Also DWORD is a windows thing so looked odd in shared code and not sure if it matches what windows thinks DWORD is.

---------
### Progress

### Reviewers
 * [Robbin Ehn](https://openjdk.java.net/census#rehn) (@robehn - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/122/head:pull/122` \
`$ git checkout pull/122`

`$ git checkout pull/122` \
`$ git pull https://git.openjdk.java.net/loom pull/122/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 122`

`$ git pr show -t 122`

</details>

<a href="https://git.openjdk.java.net/loom/pull/122.diff">https://git.openjdk.java.net/loom/pull/122.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/runtime/continuation.cpp
+++ b/src/hotspot/share/runtime/continuation.cpp
@@ -174,14 +174,15 @@ Address |   |                            |    |   Caller is still in the chunk.
 #ifdef ASSERT
 extern "C" bool dbg_is_safe(const void* p, intptr_t errvalue); // address p is readable and *(intptr_t*)p != errvalue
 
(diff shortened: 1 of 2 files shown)


robehn (approved):

coleenp:
Thanks Robbin.
/integrate

PR #128, opened 2022-03-31
Invoke implDeregister() at wakeup()

miao-zheng:
A virtual thread may park itself and register an event of EPoll when it invoke an operation of NioSocket. If the Inputstream or NioSocket closed, the call stack is like:
```
Poller::wakeup
Poller::stopPoll
NioSocketImpl.close()
···
```

I have a question that do we need invoke implDeregister() before unpark virtual thread?

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/128/head:pull/128` \
`$ git checkout pull/128`

`$ git checkout pull/128` \
`$ git pull https://git.openjdk.java.net/loom pull/128/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 128`

`$ git pr show -t 128`

</details>

<a href="https://git.openjdk.java.net/loom/pull/128.diff">https://git.openjdk.java.net/loom/pull/128.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/sun/nio/ch/Poller.java
+++ b/src/java.base/share/classes/sun/nio/ch/Poller.java
@@ -336,6 +336,7 @@ public abstract class Poller {
     private void wakeup(int fdVal) {
         Thread t = map.remove(fdVal);
         if (t != null) {
+            implDeregister(fdVal);
             LockSupport.unpark(t);
         }
     }
@@ -344,7 +345,10 @@ public abstract class Poller {
      * Called by the polling facility when the file descriptor is polled
      */
     final void polled(int fdVal) {
-        wakeup(fdVal);
+        Thread t = map.remove(fdVal);
+        if (t != null) {
+            LockSupport.unpark(t);
+        }
     }
 
     /**


joeyleeeeeee97:
Hi, I think  `implDeregister` is unnecessary now because all Poller implementation are registering with ONE SHOT options like `EPOLLONESHOT`, which will automatically remove fd once events received.

miao-zheng:
@AlanBateman Thanks for your review. If we do not explicit deregister will result epoll monitor a closed fd. Because a closed fd can be reused, I have not construct a case that the number of fd monitor by epoll overflow.

AlanBateman:
> @AlanBateman Thanks for your review. If we do not explicit deregister will result epoll monitor a closed fd. Because a closed fd can be reused, I have not construct a case that the number of fd monitor by epoll overflow.

Would it be possible to provide a summary of the scenario that you are concerned about? It's one shot so will be disarmed when polled. If the thread is interrupted or unparked for some other reason then it will deregister the file descriptor. If the socket is closed then it will be removed the polling mechanism (the dup2 trick isn't not used with virtual threads, in case that is what you mean).

miao-zheng:
@AlanBateman Thanks for your review. The scenario is:
1. A virtual thread invoke InputStream.read();
2. Virtual thread register a fd into polling mechanism and park itself;
3. Another thread invoke InputStream.close(), and it will invoke wakeup() finally;
4. The fd which is registered at InputStream.read() will deregister at polling mechanism but not Deregister at epoll, so it will monitor by epoll.

I agree that we do not explicitly invoke implDeregister is ok, so I close this PR.

miao-zheng:
@AlanBateman Thanks for your review. The scenario is:
1. A virtual thread invoke InputStream.read();
2. Virtual thread register a fd into polling mechanism and park itself;
3. Another thread invoke InputStream.close(), and it will invoke wakeup() finally;
4. the map will remove the closed fd at wakeup(), and unpark vt which is invoke InputStream.read();
5. the unparked vt will invoke deregister() and it will find "null" at map.remove();

The result is ,the fd which is registered at InputStream.read() will remove at the map which holds the fd and corresponding parked vt, but not Deregister at epoll, so it will still monitor by epoll.

As your say, there are there cases:
1. polled: automatically disarmed
2. the thread is interrupted or unparked for some other reason: remove the fd at map and invoke ImplDeregister() of Epoll
3. the socket is closed: only remove the fd at map

I think we should do the same operation of 2 and 3, and I agree that we do not explicitly invoke implDeregister is ok, so I close this PR.

PR #139, opened 2022-04-05
Some more descriptive names, removed redundant asserts.  Request for comments.

coleenp:
Also, there's an if statement already for chunk_available which the compiler shouldn't have trouble optimizing.  Otherwise we should refactor freeze_fast.
Tested locally with make test TEST=jdk/jdk/internal/vm/Continuation since only names.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/139/head:pull/139` \
`$ git checkout pull/139`

`$ git checkout pull/139` \
`$ git pull https://git.openjdk.java.net/loom pull/139/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 139`

`$ git pr show -t 139`

</details>

<a href="https://git.openjdk.java.net/loom/pull/139.diff">https://git.openjdk.java.net/loom/pull/139.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/continuation_aarch64.inline.hpp
+++ b/src/hotspot/cpu/aarch64/continuation_aarch64.inline.hpp
@@ -86,9 +86,8 @@ frame FreezeBase::new_heap_frame(frame& f, frame& caller) {
     *hf.addr_at(frame::interpreter_frame_locals_offset) = frame::sender_sp_offset + locals - 1;
     return hf;
   } else {
-    // we need to re-read fp because it may be an oop and we might have had a safepoint in finalize_freeze,
-    // after constructing f.
-    // This comment doesn't make sense since we don't reread fp
+    // We need to re-read fp out of the frame because it may be an oop and we might have
+    // had a safepoint in finalize_freeze, after constructing f.
     fp = *(intptr_t**)(f.sp() - frame::sender_sp_offset);
 
     int fsize = FKind::size(f);
--- a/src/hotspot/cpu/aarch64/continuation_aarch64.inline.hpp
+++ b/src/hotspot/cpu/aarch64/continuation_aarch64.inline.hpp
@@ -34,9 +34,9 @@
 
 // Fast path
 
-inline void FreezeBase::patch_chunk_pd(intptr_t* vsp, intptr_t* hsp) {
+inline void FreezeBase::patch_chunk_pd(intptr_t* frame_sp, intptr_t* heap_sp) {
   // copy the spilled fp from the heap to the stack
-  *(vsp - frame::sender_sp_offset) = *(hsp - frame::sender_sp_offset);
+  *(frame_sp - frame::sender_sp_offset) = *(heap_sp - frame::sender_sp_offset);
 }
 
 // Slow path
@@ -192,19 +192,19 @@ template<typename FKind> frame ThawBase::new_stack_frame(const frame& hf, frame&
   assert(FKind::is_instance(hf), "");
 
   if (FKind::interpreted) {
-    intptr_t* hsp = hf.unextended_sp();
+    intptr_t* heap_sp = hf.unextended_sp();
     const int fsize = ContinuationHelper::InterpretedFrame::frame_bottom(hf) - hf.unextended_sp();
     const int locals = hf.interpreter_frame_method()->max_locals();
-    intptr_t* vsp = caller.unextended_sp() - fsize;
-    intptr_t* fp = vsp + (hf.fp() - hsp);
+    intptr_t* frame_sp = caller.unextended_sp() - fsize;
+    intptr_t* fp = frame_sp + (hf.fp() - heap_sp);
     if ((intptr_t)fp % frame::frame_alignment != 0) {
       fp--;
-      vsp--;
+      frame_sp--;
     }
     DEBUG_ONLY(intptr_t* unextended_sp = fp + *hf.addr_at(frame::interpreter_frame_last_sp_offset);)
-    assert(vsp == unextended_sp, "");
+    assert(frame_sp == unextended_sp, "");
     caller.set_sp(fp + frame::sender_sp_offset);
-    frame f(vsp, vsp, fp, hf.pc());
+    frame f(frame_sp, frame_sp, fp, hf.pc());
     // it's set again later in derelativize_interpreted_frame_metadata, but we need to set the locals now so that we'll have the frame's bottom
     intptr_t offset = *hf.addr_at(frame::interpreter_frame_locals_offset);
     assert((int)offset == locals + frame::sender_sp_offset - 1, "");
@@ -213,16 +213,16 @@ template<typename FKind> frame ThawBase::new_stack_frame(const frame& hf, frame&
     return f;
   } else {
     int fsize = FKind::size(hf);
-    intptr_t* vsp = caller.unextended_sp() - fsize;
+    intptr_t* frame_sp = caller.unextended_sp() - fsize;
     if (bottom || caller.is_interpreted_frame()) {
       int argsize = hf.compiled_frame_stack_argsize();
 
       fsize += argsize;
-      vsp   -= argsize;
+      frame_sp   -= argsize;
       caller.set_sp(caller.sp() - argsize);
-      assert(caller.sp() == vsp + (fsize-argsize), "");
+      assert(caller.sp() == frame_sp + (fsize-argsize), "");
 
-      vsp = align(hf, vsp, caller, bottom);
+      frame_sp = align(hf, frame_sp, caller, bottom);
     }
 
     assert(hf.cb() != nullptr, "");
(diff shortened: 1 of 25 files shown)


src/hotspot/cpu/x86/continuation_x86.inline.hpp
```diff
@@ -99,7 +99,11 @@ frame FreezeBase::new_hframe(frame& f, frame& caller) {
     *hf.addr_at(frame::interpreter_frame_locals_offset) = frame::sender_sp_offset + locals - 1;
     return hf;
   } else {
-    fp = *(intptr_t**)(f.sp() - frame::sender_sp_offset); // we need to re-read fp because it may be an oop and we might have had a safepoint in finalize_freeze, after constructing f.
+    // we need to re-read fp because it may be an oop and we might have had a safepoint in
+    // finalize_freeze, after constructing f.
+    // This comment doesn't make sense since we don't reread fp
```
pron, line 104: We read the spilled fp from the frame into `fp`.

src/hotspot/cpu/aarch64/continuation_aarch64.inline.hpp
```diff
@@ -85,7 +85,11 @@ frame FreezeBase::new_hframe(frame& f, frame& caller) {
     *hf.addr_at(frame::interpreter_frame_locals_offset) = frame::sender_sp_offset + locals - 1;
     return hf;
   } else {
-    fp = *(intptr_t**)(f.sp() - frame::sender_sp_offset); // we need to re-read fp because it may be an oop and we might have had a safepoint in finalize_freeze, after constructing f.
+    // we need to re-read fp because it may be an oop and we might have had a safepoint in finalize_freeze,
+    // after constructing f.
+    // This comment doesn't make sense since we don't reread fp
```
pron, line 90: See x86 response

src/hotspot/share/runtime/continuation.cpp
```diff
@@ -997,8 +997,8 @@ class Freeze : public FreezeBase {
     , int* out_size = nullptr
 #endif
   );
-  template <bool chunk_available> freeze_result try_freeze_fast(intptr_t* sp);
-  template <bool chunk_available> bool freeze_fast(intptr_t* top_sp);
+  freeze_result try_freeze_fast(intptr_t* sp, bool chunk_available);
+  bool freeze_fast(intptr_t* top_sp, bool chunk_available);
```
pron, line 1001: We must not branch on the fast path (i.e. we might have budget to branch once or maybe twice). For now, this needs to remain a template parameter. It has no significant impact on build time and is just as easy to read. This kind of thing is how we managed to avoid writing it in assembly.

src/hotspot/share/runtime/continuation.cpp
```diff
@@ -2196,30 +2195,31 @@ inline intptr_t* Thaw<ConfigT>::thaw(thaw_kind kind) {
...
 
+  // LogTarget might generate some test code
   LogTarget(Trace, continuations) lt;
   if (lt.develop_is_enabled()) {
     LogStream ls(lt);
     ls.print_cr("thaw_fast");
     chunk->print_on(true, &ls);
   }
 
+  // Comment needed: why is 500? Related to 300 in can_thaw_fast?
```
pron, line 2222: It's a heuristic. Below this number, we thaw the whole chunk. Above it, we thaw one frame. 500 seemed to work ok, but might need to be adjusted after deeper profiling.

src/hotspot/share/runtime/continuation.cpp
coleenp: This template seemed dubiously helpful and now we have 4 of this huge function, which the compiler may be able to throw away parts of.  Aside, can't see how the ConfigT parameter is used here.  Maybe we don't need that?   I'll revert it but we really dislike these templates and think that the C++ compiler can do just as good as a job optimizing this code.  Templates are not easy to read and even worse to debug, and essentially add a level of indirection for understanding the code.  The syntax and compilation messages are awful!  We need a set of benchmarks to justify this coding style.

coleenp: No, 4 is an undercount.  How many of these are there?

src/hotspot/cpu/aarch64/continuation_aarch64.inline.hpp
coleenp: ok

src/hotspot/share/runtime/continuation.cpp
```diff
@@ -2077,6 +2071,7 @@ static inline int prepare_thaw0(JavaThread* thread, bool return_barrier) {
   stackChunkOop chunk = jdk_internal_vm_Continuation::tail(continuation);
   assert(chunk != nullptr, "");
 
+  // Comment needed: Why would the tail chunk be empty? Why do you get the parent?
```
coleenp, line 2074: Here, can you explain what's happening in this function?

coleenp: Then we'll do more functions :)

src/hotspot/share/runtime/continuation.cpp
pron: It's unrelated. The 300 is just an estimate for the stack size taken by the native code. 500 is just a heuristic of when to thaw one frame or the entire chunk (so refers to Java frames). The function always makes room for the entire chunk.

src/hotspot/share/runtime/continuation.cpp
pron: We make room on the stack to thaw frames. As we'll either thaw the topmost non-empty chunk or one frame, we just need to make room for the topmost non-empty chunk.

src/hotspot/share/runtime/continuation.cpp
```diff
@@ -1557,34 +1557,43 @@ void FreezeBase::patch(const frame& f, frame& hf, const frame& caller, bool bott
...
     if (real_unextended_sp != nullptr) {
       f.set_unextended_sp(real_unextended_sp); // can be null at a safepoint
+                 // is this ever at safepoint? maybe was for preempt?
     }
   }
 #else
   Unimplemented();
 #endif
 
+  // Why 'vsp' what is 'v' ?  isn't this the sp on the real stack?
```
pron, line 1589: To distinguish between stacks, John called the continuation stack a "horizontal stack", hence `h`, and the thread stack a "vertical stack`, hence `v`.

coleenp: Ok I wouldn't have guessed.

src/hotspot/share/runtime/continuation.cpp
coleenp: So here the topmost chunk may be null.  Why is it null?  Is it null because it was already partially thawed and you leave space for a new chunk in the chunk memory?  In this case, you continue to the parent (next) frame.  Did I get that right?

src/hotspot/share/runtime/continuation.cpp
coleenp: ok.  I'll try to think of something to say.

src/hotspot/share/runtime/continuation.cpp
```diff
@@ -2087,6 +2093,8 @@ static inline int prepare_thaw0(JavaThread* thread, bool return_barrier) {
   // Verification
   chunk->verify();
 
+  // Comment needed: Just the size for the first chunk? Isn't there a linked list of chunks?
+  // Don't you need the sum of their sizes?
```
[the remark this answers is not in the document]
pron: That's correct

src/hotspot/share/runtime/continuation.cpp
```diff
@@ -2349,6 +2346,7 @@ NOINLINE intptr_t* ThawBase::thaw_slow(stackChunkOop chunk, bool return_barrier)
     chunk->print_on(true, &ls);
   }
 
+  // Does this need ifdef JFR around it? Or can we remove all the conditional JFR inclusions (better)?
```
pron, line 2359: I'll need to think about that.

src/hotspot/share/runtime/continuation.cpp
pron: Not null, but empty. I can't remember at the moment why we do it like that. I'll take a closer look.

src/hotspot/share/runtime/continuation.cpp
coleenp: s/vsp/top_sp/ or top_thread_sp ?
s/hsp/frame_sp/ ?

coleenp: s/vsp/frame_sp/
s/hsfp/heap_sp/

frame_sp can imply and looks to the reader like the normal sp in the thread stack frame.

coleenp:
Discussed with Ron and Patricio offline.  Pushing after tier1 tests pass on x64 and aarch64 platforms.
/integrate

PR #140, opened 2022-04-07
dcubed - misc cleanups of jvm.cpp and JVM/TI files.

dcubed-ojdk:
Misc cleanups of jvm.cpp and JVM/TI files.

---------
### Progress

### Reviewers
 * [Serguei Spitsyn](https://openjdk.java.net/census#sspitsyn) (@sspitsyn - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/140/head:pull/140` \
`$ git checkout pull/140`

`$ git checkout pull/140` \
`$ git pull https://git.openjdk.java.net/loom pull/140/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 140`

`$ git pr show -t 140`

</details>

<a href="https://git.openjdk.java.net/loom/pull/140.diff">https://git.openjdk.java.net/loom/pull/140.diff</a>

</details>


dcubed-ojdk:
One of the less obvious cleanups is to move the associated locals closer to
`get_threadOop_and_JavaThread()` calls to reduce the scope of the locals.

src/hotspot/share/prims/jvmti.xml
```diff
@@ -1861,10 +1861,10 @@ jvmtiEnv *jvmti;
...
-        Virtual threads may be suspended with
-        <functionlink id="SuspendAllVirtualThreads"></functionlink> or
-        <functionlink id="SuspendThreadList"></functionlink> or
-        <functionlink id="SuspendThread"></functionlink>.
+        Virtual threads may be resumed with
+        <functionlink id="ResumeAllVirtualThreads"></functionlink> or
+        <functionlink id="ResumeThreadList"></functionlink> or
+        <functionlink id="ResumeThread"></functionlink>.
         Virtual threads that are currently resumed do not change state.
       </description>
```
sspitsyn, line 1869: This is not right.
The JVM TI Suspend functions are listed intentionally. Similarly, the JVM TI Resume functions are listed in SuspendAllVirtualThreads:
http://100.110.26.5:8080/view/loom/job/loom-fibers-branch-build/lastSuccessfulBuild/artifact/loom/build/linux-x64/images/docs/specs/jvmti.html#SuspendAllVirtualThreads

src/hotspot/share/prims/jvmtiEventController.cpp
```diff
@@ -572,9 +572,9 @@ JvmtiEventControllerPrivate::recompute_thread_enabled(JvmtiThreadState *state) {
     // mark if event is truly enabled on this thread in any environment
     state->thread_event_enable()->_event_enabled.set_bits(any_env_enabled);
 
-    // pointer to JavaThread can be NULL for unmouted virtual thread
     if (state->get_thread() != NULL) {
-      // update the JavaThread cached value for thread-specific should_post_on_exceptions value
+      // Update the JavaThread or mounted virtual thread cached value for
+      // thread-specific should_post_on_exceptions value.
       bool should_post_on_exceptions = (any_env_enabled & SHOULD_POST_ON_EXCEPTIONS_BITS) != 0;
```
sspitsyn, line 578: The updated comment does not look correct, so I want to clarify. We update the JavaThread for both normal and virtual thread. However, virtual thread can be unmounted. There is no JavaThread to update in this case.
Then the updated comment (in your style) should say something like this:
```
      // The JavaThread for carrier or mounted virtual thread case.
      // Update the cached value for thread-specific should_post_on_exceptions value.
```
But I feel that the original comments were okay. :)

src/hotspot/share/prims/jvmtiThreadState.cpp
```diff
@@ -117,10 +117,10 @@ JvmtiThreadState::JvmtiThreadState(JavaThread* thread, oop thread_oop)
     _is_virtual = java_lang_VirtualThread::is_instance(thread_oop);
   }
 
-  // thread can be NULL if virtual thread is unmounted
   if (thread != NULL) {
-    // set this as the state for the thread only if thread_oop is current thread->jvmti_vthread()
     if (thread_oop == NULL || thread->jvmti_vthread() == NULL || thread->jvmti_vthread() == thread_oop) {
+      // Set this as the state for the JavaThread or mounted virtual thread
+      // only if thread_oop is current thread->jvmti_vthread().
       thread->set_jvmti_thread_state(this);
```
sspitsyn, line 124: This update has the same problem as previous one that I've commented. We set the thread->jvmti_thread_state() value in the JavaThread when it is bound to a normal/carrier or mounted virtual thread. There is no JavaThread only for unmounted virtual thread.

src/hotspot/share/prims/jvmtiThreadState.cpp
```diff
@@ -210,15 +210,17 @@ JvmtiThreadState::periodic_clean_up() {
   }
 }
 
-/* Virtual Threads Mount Transition (VTMT) mechanism */
+//
+// Virtual Threads Mount Transition (VTMT) mechanism
+//
 
-// VTMT can not be disabled while this counter is positive
+// VTMT cannot be disabled while this counter is positive
```
sspitsyn, line 217: A dot at the end is also needed to follow your comment style.

sspitsyn (approved):
Dan,
This is a great cleanup!
It is nice you used this chance to make all the comments in this area consistent.
I've posted a couple of comments which have to be easy to resolve.
Thanks,
Serguei

src/hotspot/share/prims/jvmti.xml
```diff
@@ -1861,10 +1861,10 @@ jvmtiEnv *jvmti;
...
         Resume all virtual threads except those already resumed.
-        Virtual threads may be suspended with
-        <functionlink id="SuspendAllVirtualThreads"></functionlink> or
-        <functionlink id="SuspendThreadList"></functionlink> or
-        <functionlink id="SuspendThread"></functionlink>.
+        Virtual threads may be resumed with
+        <functionlink id="ResumeAllVirtualThreads"></functionlink> or
+        <functionlink id="ResumeThreadList"></functionlink> or
+        <functionlink id="ResumeThread"></functionlink>.
         Virtual threads that are currently resumed do not change state.
```
AlanBateman, line 1868: Just to add to Serguei's comment. I think it would be clearer if we move the sentence "Virtual threads that are currently resumed do not change state" to before the "Virtual threads may be suspended ..."

dcubed-ojdk: I moved the sentence as @AlanBateman suggested.
I also moved the equivalent sentence in SuspendAllVirtualThreads.

src/hotspot/share/prims/jvmti.xml
dcubed-ojdk: I have reverted this change. Sorry for the noise.

src/hotspot/share/prims/jvmtiEventController.cpp
dcubed-ojdk: I've updated to use the revised comment you proposed.

src/hotspot/share/prims/jvmtiThreadState.cpp
dcubed-ojdk: I've updated the comment to:
```
      // The JavaThread for carrier or mounted virtual thread case.
      // Set this only if thread_oop is current thread->jvmti_vthread().
```
to try and match your style from your previous comment. Please let me know
if the new version works for you.

src/hotspot/share/prims/jvmtiThreadState.cpp
dcubed-ojdk: Fixed.

sspitsyn (approved):
Thank you for addressing the comments!
It looks good now.
Thanks,
Serguei

dcubed-ojdk:
@sspitsyn - Thanks the review and the re-review.
@AlanBateman - Thanks for chiming in on the thread.

I don't know if this will work, but here goes:
/integrate

PR #142, opened 2022-04-11
Minor changes to `StructuredTaskScope`

PaulSandoz:
Some suggested minor changes to `StructuredTaskScope` source while reviewing the code.

Overall I like how this approach has boiled down to about as simple as it can get while supporting the constraints of structured concurrency (thanks in no small part to `ThreadFlock`).

I think it should be possible to simplify the subclass implementations if the values of `Future.State` were comparable as in `RUNNING` < `CANCELED` < `FAILED` < `SUCCESS`. Otherwise, introducing that internally could also simplify and avoid the task scopes holding on unnecessarily to futures that are not operated on after join. I was reluctant to make such a change here but could propose as a separate PR.

I was uncertain if the `ShutdownOnSuccess.result` methods and the various exception returning/operating methods on `ShutdownOnFailure` should be constrained to throw if called before `join` has completed. Otherwise, we should specify that if operated on before `join` has completed then the results are unspecified.
(There might be a simple way for join to returning something other than `this`, thereby avoiding this issue, with some additional complexity, but I suspect you have already thought of that and took the minimal route.)

I presume there is an implicit happens-before edge on `join` that we can make explicit and specify?

### DIFF
--- a/src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/StructuredTaskScope.java
+++ b/src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/StructuredTaskScope.java
@@ -61,7 +61,7 @@ import jdk.internal.misc.ThreadFlock;
  * A StructuredTaskScope is created with one of its public constructors. It defines
  * the {@link #fork(Callable) fork} method to start a thread to execute a task, the {@link
  * #join() join} method to wait for all threads to finish, and the {@link #close() close}
-  * method to close the task scope. The API is intended to be used with the {@code
+ * method to close the task scope. The API is intended to be used with the {@code
  * try-with-resources} construct. The intention is that code in the <em>block</em> uses
  * the {@code fork} method to fork threads to execute the sub-tasks, wait for the threads
  * to finish with the {@code join} method, and then <em>process the results</em>.
@@ -90,9 +90,9 @@ import jdk.internal.misc.ThreadFlock;
  * {@code join} to wakeup. It also interrupts all unfinished threads and prevents new threads
  * from starting in the task scope.
  *
- * <h2>Sub-classes with policies for common cases</h2>
+ * <h2>Subclasses with policies for common cases</h2>
  *
- * Two sub-classes of StructuredTaskScope are defined to implement policy for common cases:
+ * Two subclasses of StructuredTaskScope are defined to implement policy for common cases:
  * <ol>
  *   <li> {@link ShutdownOnSuccess ShutdownOnSuccess} captures the first result and
  *   shuts down the task scope to interrupt unfinished threads and wakeup the owner. This class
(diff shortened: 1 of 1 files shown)


pron:
Thank you!
We've explored policies different `join` signatures wrapping the STS class rather than extending it. Ultimately, we favoured a common supertype with a common `join` method that better exposes the similarities between them, and `join` returning `this` is a reasonable compromise between expressing the similarity and a convenient, typesafe API.

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/StructuredTaskScope.java
```diff
@@ -269,17 +269,21 @@
     private final ThreadFlock flock;
     private final ReentrantLock shutdownLock = new ReentrantLock();
 
-    // the set of "tracked" Future objects, created lazily
+    // the set of "tracked" Future objects waiting to be returned by Future.get, created lazily
+    // assigned once-only with FUTURES.compareAndSet, read by any thread
```
AlanBateman, line 273: Instead of "assigned once-only ..." it might be simpler to just link to the track method

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/StructuredTaskScope.java
```diff
@@ -342,6 +346,7 @@ private boolean isShutdown() {
      */
     private void track(Future<?> future) {
         // create the set of Futures if not already created
+        // cannot use double-checked locking since it requires use of a synchronized block
```
AlanBateman, line 349: I'd prefer to leave this comment as is, only because mentioning another possible implementation might confuse reader.

AlanBateman (commented):
Sub-class -> subclass, shutdown -> shut down, and moving the unchecked cast to FutureImpl are good - thanks. Just a few comments on the changes to the internal comments.

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/StructuredTaskScope.java
PaulSandoz: Ok, removed. When/if monitors are supported properly in loom we can use double checked locking, which is more idiomatic. Hard to track these things.

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/StructuredTaskScope.java
PaulSandoz: Changed to "    // assigned to non-null value in method track, read by any thread"

PaulSandoz:
I added a class doc section on memory consistency effects, and updated the access methods on the subclasses. 

I propose another PR, once this one has integrated, on the `ShutdownOnSuccess` and `ShutdownOnFailure` implementations to contrast and lets see if its useful.

AlanBateman (approved):

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/StructuredTaskScope.java
```diff
@@ -251,6 +251,14 @@
  * <p> Unless otherwise specified, passing a {@code null} argument to a constructor
  * or method in this class will cause a {@link NullPointerException} to be thrown.
  *
+ * <h2>Memory consistency effects</h2>
+ * <p>Actions in the owner thread of, or a thread contained in, the task scope prior to
+ * {@linkplain #fork forking} of a {@code Runnable} or {@code Callable} task
```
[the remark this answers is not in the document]
PaulSandoz: Doh! thanks. I also fixed the link.

PR #144, opened 2022-04-13
Use one future

PaulSandoz:
Here's an example of changing the structured task scope impls, using one `Future` field and comparing future states. In effect the state can only increase monotonically (as per the order determined by the comparator).

The single future field could be made non-volatile, using the `VarHandle` for the volatile read in `handleComplete` since the happens-before edge should ensure the future is visible to the reading thread after `join`.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/144/head:pull/144` \
`$ git checkout pull/144`

`$ git checkout pull/144` \
`$ git pull https://git.openjdk.java.net/loom pull/144/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 144`

`$ git pr show -t 144`

</details>

<a href="https://git.openjdk.java.net/loom/pull/144.diff">https://git.openjdk.java.net/loom/pull/144.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/StructuredTaskScope.java
+++ b/src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/StructuredTaskScope.java
@@ -31,6 +31,7 @@ import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.Comparator;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/StructuredTaskScope.java
```diff
@@ -797,6 +798,19 @@ public String toString() {
         }
     }
 
+    private static int stateToId(Future.State s) {
```
AlanBateman, line 801: Did you mean to to name this stateToId or stateToInt? 
Style-wise I guess I'd would align the arrows but it doesn't matter.

AlanBateman (approved):
Clever!

AlanBateman (approved):

PR #145, opened 2022-04-14
Split out ContinuationEntry and ContinuationWrapper classes

coleenp:
Moved ContinuationEntry and ContinuationWrapper to their own files.
There's a bit of inconsistency because 3 ContinuationEntry functions are in ContinuationHelper_cpu.inline.hpp - maybe we should have continuationEntry_cpu.inline.hpp files for just the three.
I'd like to change the name ContinuationEntry to ContinuationAnchor or ContinuationFrameAnchor but want to study how that would look a bit more first.
Tested with other platforms and testing with loom-tier1 as sanity.

---------
### Progress

### Reviewers
 * [Ron Pressler](https://openjdk.java.net/census#rpressler) (@pron - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/145/head:pull/145` \
`$ git checkout pull/145`

`$ git checkout pull/145` \
`$ git pull https://git.openjdk.java.net/loom pull/145/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 145`

`$ git pr show -t 145`

</details>

<a href="https://git.openjdk.java.net/loom/pull/145.diff">https://git.openjdk.java.net/loom/pull/145.diff</a>

</details>


src/hotspot/share/runtime/continuationWrapper.cpp
```diff
@@ -0,0 +1,135 @@
...
+  : _thread(thread), _entry(thread->last_continuation()), _continuation(continuation),
+    _e_size(0), _e_num_interpreted_frames(0)
+  {
+  assert(oopDesc::is_oop(_continuation),
+         "Invalid continuation object: " INTPTR_FORMAT, p2i((void*)_continuation));
+  assert(_continuation == _entry->cont_oop(), "cont: " INTPTR_FORMAT " entry: " INTPTR_FORMAT " entry_sp: "
+         INTPTR_FORMAT, p2i((oopDesc*)_continuation), p2i((oopDesc*)_entry->cont_oop()), p2i(entrySP()));
+  disallow_safepoint();
+  read();
+}
```
pron, line 50: This constructor should go in the .inline file, and the intialization of the JFR fields should be #ifdefed

src/hotspot/share/runtime/continuationWrapper.inline.hpp
```diff
@@ -46,10 +47,8 @@ class ContinuationWrapper : public StackObj {
   oop                _continuation;  // jdk.internal.vm.Continuation instance
   stackChunkOop      _tail;
 
-#if CONT_JFR // Profiling data for the JFR event
   short _e_size;
   short _e_num_interpreted_frames;
```
pron, line 51: The decalration of these two fields needs to be guarded by `#if CONT_JFR`.

pron:
Because the JFR event fields are only used in freeze and thaw (and only for perf debugging), they could go into a common superclass (FreezeThawBase) in continuationFreezeThaw.cpp and removed from ContinuationWrapper altogether.

stefank (changes requested):
Thanks for splitting the files. I agree that it would make sense to move out the ContinuationEntry functions out of the ContinuationHelper files. I hope you get to that after this change.

There are number of inconsitencies in whitespace, include guards, and include orders. I've fixed those as a patch on top of your changes:
https://github.com/openjdk/loom/compare/fibers...stefank:pr_loom_145

Would you mind including those in your PR?

Thanks

src/hotspot/share/runtime/continuationWrapper.cpp
```diff
@@ -0,0 +1,135 @@
...
+  for (stackChunkOop chunk = tail(); chunk != nullptr; chunk = chunk->parent()) {
+    if (chunk->is_in_chunk(p)) {
+      assert(chunk->is_usable_in_chunk(p), "");
+      return chunk;
+    }
+  }
+  return nullptr;
+}
+
+template<typename Event> void ContinuationWrapper::post_jfr_event(Event* e, JavaThread* jt) {
```
pron, line 94: This method can be moved into FreezeThawBase in continuationFreezeThaw.cpp.

src/hotspot/share/runtime/continuationWrapper.inline.hpp
coleenp: I don't see the point of having this mess for two variable initializations.

src/hotspot/share/runtime/continuationWrapper.cpp
coleenp: I thought I should move these back.

src/hotspot/share/runtime/continuationWrapper.inline.hpp
coleenp: maybe all the JFR code can go in freezethaw as you suggest.

coleenp: I moved the jfr code in continuationFreezeThaw.cpp

pron (approved):

pron (approved):

coleenp:
Thanks Ron.  I'm waiting for tier1 to run everywhere first.

coleenp:
Most of the tests ran anyway.
/integrate

PR #146, opened 2022-04-14
Debugger changes to support JVMTI now returning JVMTI_ERROR_UNSUPPORTED_OPERATION for Thread.stop() of a virtual thread.

plummercj:
---------
### Progress

### Reviewers
 * [Serguei Spitsyn](https://openjdk.java.net/census#sspitsyn) (@sspitsyn - Committer) ⚠️ Review applies to 73e710eacb3c2ff1d20914d461ea6274a670da42

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/146/head:pull/146` \
`$ git checkout pull/146`

`$ git checkout pull/146` \
`$ git pull https://git.openjdk.java.net/loom pull/146/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 146`

`$ git pr show -t 146`

</details>

<a href="https://git.openjdk.java.net/loom/pull/146.diff">https://git.openjdk.java.net/loom/pull/146.diff</a>

</details>


test/hotspot/jtreg/ProblemList-svc-vthread.txt
```diff
@@ -24,8 +24,6 @@
 ####
 # Bugs
 
-vmTestbase/nsk/jdb/kill/kill001/kill001.java                00000000 generic-all
-
```
sspitsyn, line 28: This test was problem listed on mac-osx (the reason was this bug https://bugs.openjdk.java.net/browse/JDK-8282383) before Leonid problem listed it generally. So we want to restore this? Also, this test is failing with timeout when I run it locally on linux-x64-debug.

sspitsyn (approved):
I've posted one comment.
Otherwise, it looks okay to me.
Also, the debugger needs to be tested with mach5 on all platforms.
Thanks,
Serguei

src/java.se/share/data/jdwp/jdwp.spec
```diff
@@ -2021,8 +2021,9 @@ JDWP "Java(tm) Debug Wire Protocol"
...
         )
         (ErrorSet
-            (Error INVALID_THREAD "The thread is null, not a valid thread, the thread "
-                                  "is not alive, or the thread is a virtual thread "
+            (Error INVALID_THREAD "The thread is null, not a valid thread, or the thread "
+                                  "is not alive.")
+            (Error NOT_IMPLEMENTED "The thread is a virtual thread "
                                   "and the target VM does not support stopping it "
                                   "with an asynchronous exception at this time.")
             (Error INVALID_OBJECT "If thread is not a known ID or the asynchronous "
```
AlanBateman, line 2029: I think we can reduce this down to "The thread is a virtual thread and the target VM does not support the operation".   The phrase "at this time" worked when it make have been dependent on whether the thread was at a breakpoint (or even the thread state) but I think it needs to be dropped now.

src/jdk.jdi/share/classes/com/sun/jdi/ThreadReference.java
```diff
@@ -125,9 +125,10 @@ public interface ThreadReference extends ObjectReference {
      * @param throwable the asynchronous exception to throw
      * @throws InvalidTypeException if <code>throwable</code> is not
      * an instance of java.lang.Throwable in the target VM
-     * @throws IllegalThreadStateException if the thread has terminated,
-     * or the thread is a virtual thread that cannot be stopped with an
-     * asynchronous exception at this time
+     * @throws IllegalThreadStateException if the thread has terminated
+     * @throws java.lang.UnsupportedOperationException if the thread is a
+     * virtual thread that cannot be stopped with an asynchronous exception
```
AlanBateman, line 130: Minor comment is that you can drop java.lang if you want as that packaged is always imported.

src/java.se/share/data/jdwp/jdwp.spec
```diff
@@ -2021,10 +2021,10 @@ JDWP "Java(tm) Debug Wire Protocol"
...
         )
         (ErrorSet
-            (Error INVALID_THREAD "The thread is null, not a valid thread, the thread "
-                                  "is not alive, or the thread is a virtual thread "
-                                  "and the target VM does not support stopping it "
-                                  "with an asynchronous exception at this time.")
+            (Error INVALID_THREAD "The thread is null, not a valid thread, or the thread "
+                                  "is not alive.")
+            (Error NOT_IMPLEMENTED "The thread is a virtual thread and the target VM "
+                                  "VM does not support the operation.")
```
AlanBateman, line 2027: The updated version repeats "VM".

plummercj: Oops. Already integrated. Will do a direct push to fix. Also, I think the JDI wording should be made to look similar.

plummercj: I also think it should be "command" instead of "operation" to be consistent with the text a little bit above this that reads:

        "The target VM may not support, or may only provide limited support, for "
        "this command when the thread is a virtual thread. It may, for example, "

AlanBateman: > I also think it should be "command" instead of "operation" to be consistent with the text a little bit above this that reads:

Okay, although the resulting "The thread is a virtual thread and the target VM does not support the command" should probably be expanded to make it clear that the command is not supported on virtual threads. This would avoid an interpreting it to mean that the error is returned when the VM doesn't support the command AND the thread ID is for a virtual thread.

plummercj: https://github.com/openjdk/loom/commit/5a2fbb7438e551ffbecedb5530d08bae5d79b74d

I tried to be consistent with the descriptive text above the error/exception section, thus "the command" for JDWP and "this operation" for JDI.

PR #149, opened 2022-04-15
Alex Menkov review

plummercj:
Minor fixes based on input from Alex Menkov

---------
### Progress

### Reviewers
 * [Alex Menkov](https://openjdk.java.net/census#amenkov) (@alexmenkov - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/149/head:pull/149` \
`$ git checkout pull/149`

`$ git checkout pull/149` \
`$ git pull https://git.openjdk.java.net/loom pull/149/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 149`

`$ git pr show -t 149`

</details>

<a href="https://git.openjdk.java.net/loom/pull/149.diff">https://git.openjdk.java.net/loom/pull/149.diff</a>

</details>


src/java.se/share/data/jdwp/jdwp.spec
```diff
@@ -2149,10 +2149,10 @@ JDWP "Java(tm) Debug Wire Protocol"
...
             (Error OPAQUE_FRAME      "Attempted to return early from "
                                      "a frame corresponding to a native "
-                                     "method. "
-                                     "The thread is a virtual thread and the target "
-                                     "VM is unable force its current frame to return. "
-                                     "Or the implementation is "
+                                     "method, "
+                                     "or the thread is a virtual thread and the target "
+                                     "VM is unable to force its current frame to return, "
+                                     "or the implementation is "
```
AlanBateman, line 2155: It's okay to join the sentences but if you do then we need to re-forward L2152 so that we don't have a line break after the comma at L2152.

src/jdk.jdi/share/classes/com/sun/jdi/request/ThreadStartRequest.java
```diff
@@ -53,8 +53,6 @@ public interface ThreadStartRequest extends EventRequest {
     /**
      * Restricts the events generated by this request to those in
      * the given thread.
-     * The behavior of this method is unspecified when the event is restricted
-     * to only platform threads.
      * @param thread the thread to filter on.
```
AlanBateman, line 56: Dropped the "The behavior ..." from both requests and in both methods is okay.

src/java.se/share/data/jdwp/jdwp.spec
plummercj: I don't understand. Aren't they all formatted into one long line, and line wrapping is controlled by how the browser formats the table cell?

https://docs.oracle.com/en/java/javase/17/docs/specs/jdwp/jdwp-protocol.html#JDWP_ThreadReference_ForceEarlyReturn

AlanBateman: Yes, that's the generated spec. The comment was on the source file which can be cleaned up, or later if you want, it doesn't matter.

plummercj: Ok, so you just want the spec to be better formatted. I can fix that now.

src/java.se/share/data/jdwp/jdwp.spec
```diff
@@ -2147,14 +2147,12 @@ JDWP "Java(tm) Debug Wire Protocol"
...
-                                     "VM is unable force its current frame to return. "
-                                     "Or the implementation is "
-                                     "unable to provide this functionality "
-                                     "on this frame.")
+            (Error OPAQUE_FRAME      "Attempted to return early from a frame "
+                                     "corresponding to a native method, "
+                                     "or the thread is a virtual thread and the target "
+                                     "VM is unable to force its current frame to return, "
+                                     "or the implementation is unable to provide this "
+                                     "functionality on this frame.")
```
AlanBateman, line 2155: No objection to combined the sentences but there is an order "or" and a missing comma. I think this is how it should read in list form:

"Attempted to return early from a frame corresponding to a native method, the thread is a virtual thread and the target VM is unable to force its current frame to return, or the implementation is unable to provide this functionality on this frame."

AlanBateman (approved):
Thanks for update.

alexmenkov (approved):

PR #156, opened 2022-04-19
Freeze functions

robehn:
---------
### Progress

### Reviewers
 * [Coleen Phillimore](https://openjdk.java.net/census#coleenp) (@coleenp - Committer) ⚠️ Review applies to d75f034f8c45b7ed573f8d2eba3a0f189230cb61

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/156/head:pull/156` \
`$ git checkout pull/156`

`$ git checkout pull/156` \
`$ git pull https://git.openjdk.java.net/loom pull/156/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 156`

`$ git pr show -t 156`

</details>

<a href="https://git.openjdk.java.net/loom/pull/156.diff">https://git.openjdk.java.net/loom/pull/156.diff</a>

</details>

### DIFF
--- a/test/hotspot/jtreg/serviceability/jvmti/HeapMonitor/MyPackage/HeapMonitorVMEventsTest.java
+++ b/test/hotspot/jtreg/serviceability/jvmti/HeapMonitor/MyPackage/HeapMonitorVMEventsTest.java
@@ -26,6 +26,7 @@ package MyPackage;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ThreadFactory;
 
 // Graal is not tested here due to Graal not supporting DisableIntrinsic.
 /**
@@ -34,9 +35,8 @@ import java.util.List;
  * @requires vm.jvmti
  * @requires !vm.graal.enabled
  * @build Frame HeapMonitor
- * @compile --enable-preview -source ${jdk.version} HeapMonitorVMEventsTest.java
- * @run main/othervm/native --enable-preview
- *                          -XX:+UnlockDiagnosticVMOptions
+ * @compile HeapMonitorVMEventsTest.java
+ * @run main/othervm/native -XX:+UnlockDiagnosticVMOptions
  *                          -XX:DisableIntrinsic=_clone
  *                          -agentlib:HeapMonitorTest MyPackage.HeapMonitorVMEventsTest platform
  */
@@ -116,12 +116,24 @@ public class HeapMonitorVMEventsTest implements Cloneable {
     HeapMonitor.sampleEverything();
 
     if(args[0].equals("virtual")) {
-        Thread t = Thread.ofVirtual().start(() -> {
-                compareSampledAndVM();
-            });
+        Thread t = virtualThreadFactory().newThread(HeapMonitorVMEventsTest::compareSampledAndVM);
+        t.start();
         t.join();
     } else {
         compareSampledAndVM();
+    }
+  }
+
+  private static ThreadFactory virtualThreadFactory() {
+    try {
+        Object builder = Thread.class.getMethod("ofVirtual").invoke(null);
+        Class<?> clazz = Class.forName("java.lang.Thread$Builder");
+        java.lang.reflect.Method factory = clazz.getMethod("factory");
+        return (ThreadFactory) factory.invoke(builder);
+    } catch (RuntimeException | Error e) {
+        throw e;
+    } catch (Exception e) {
+        throw new RuntimeException(e);
     }
   }
 }


src/hotspot/share/runtime/continuationFreezeThaw.cpp
```diff
@@ -333,30 +333,45 @@ class FreezeBase : public StackObj {
...
 
-  intptr_t *_bottom_address;
+  intptr_t* _bottom_address;
 
   int _size; // total size of all frames plus metadata in words.
   int _align_size;
 
+  intptr_t* _cont_stack_top;
+  intptr_t* _cont_stack_bottom;
+  int _cont_size;
```
coleenp, line 347: I know these are in logical grouping and it's a StackObj, but can you move the two bools to after cont_size to minimize alignment gaps.  Also there are three sizes now in this class.  Can we have bigger comments for what each means in the declaration?  I think this will be helpful for the abstraction.

src/hotspot/share/runtime/continuationFreezeThaw.cpp
```diff
@@ -333,30 +333,45 @@ class FreezeBase : public StackObj {
 protected:
   JavaThread* const _thread;
   ContinuationWrapper& _cont;
-  CONT_JFR_ONLY(FreezeThawJfrInfo _jfr_info;)
   bool _barriers;
   const bool _preempt; // used only on the slow path
+  const intptr_t * const _frame_sp;
```
coleenp, line 338: Is this the sp of the first frame on the stack that we are going to freeze?

src/hotspot/share/runtime/continuationFreezeThaw.cpp
```diff
@@ -473,9 +500,15 @@ void FreezeBase::unwind_frames() {
...
 template <bool chunk_available>
-freeze_result Freeze<ConfigT>::try_freeze_fast(intptr_t* sp) {
-  if (freeze_fast<chunk_available>(sp)) {
+freeze_result Freeze<ConfigT>::try_freeze_fast() {
+  if (chunk_available) {
+    freeze_fast_existing_chunk();
     return freeze_ok;
+  } else {
+    stackChunkOop chunk = allocate_chunk(_cont_size + frame::metadata_words);
+    if (freeze_fast_new_chunk(chunk)) {
```
coleenp, line 509: It seems odd that freeze_fast_existing_chunk() doesn't pass a chunk but freeze_fast_new_chunk does.  Should the allocate_chunk be inside of freeze_fast_new_chunk?

src/hotspot/share/runtime/continuationFreezeThaw.cpp
```diff
@@ -507,108 +535,93 @@ bool Freeze<ConfigT>::is_chunk_available_for_fast_freeze(intptr_t* frame_sp
 
   // assert(CodeCache::find_blob(*(address*)(frame_sp - SENDER_SP_RET_ADDRESS_OFFSET)) == StubRoutines::cont_doYield_stub(), ""); -- fails on Windows
   assert(StubRoutines::cont_doYield_stub()->frame_size() == frame::metadata_words, "");
-  intptr_t* const stack_top     = frame_sp + frame::metadata_words;
-  intptr_t* const stack_bottom  = _cont.entrySP() - ContinuationHelper::frame_align_words(_cont.argsize());
-
-  int size = stack_bottom - stack_top; // in words
 
+  int ret_size = _cont_size;
```
coleenp, line 539: ret_size is the total size that you need to freeze right?  Maybe call it total_size_needed or something better.

src/hotspot/share/runtime/continuationFreezeThaw.cpp
```diff
@@ -507,108 +535,93 @@ bool Freeze<ConfigT>::is_chunk_available_for_fast_freeze(intptr_t* frame_sp
...
+  if (UNLIKELY(chunk == nullptr || !_thread->cont_fastpath() || _barriers)) { // OOME/probably humongous
+    log_develop_trace(continuations)("Retrying slow. Barriers: %d", _barriers);
+    return false;
+  }
 
-    // in a fresh chunk, we freeze *with* the bottom-most frame's stack arguments.
-    // They'll then be stored twice: in the chunk and in the parent chunk's top frame
-    chunk_start_sp = cont_size + frame::metadata_words;
-    assert(chunk_start_sp == chunk->stack_size(), "");
+  chunk->set_max_size(_cont_size);
```
coleenp, line 608: This variable _cont_size is really strange.  Above we add in argsize for total size that we need to freeze, but here we set max_size to cont_size.  We should go through and see what we add and subtract from cont_size in various places and see what size is what (and if it's easier to figure out what's correct).  After this patch.
I assume that this patch is mostly code motion and doesn't change the values in places, which is good.

coleenp (approved):
Hard to tell with diffs but I think this looks good.  I like the refactoring of freeze_fast_copy.

src/hotspot/share/runtime/continuationFreezeThaw.cpp
robehn: Yes

src/hotspot/share/runtime/continuationFreezeThaw.cpp
robehn: The chunk is picked up from tail. Yes I agree it's not the way I liked.
But want to ship smaller pieces, since merging is really rough.

src/hotspot/share/runtime/continuationFreezeThaw.cpp
robehn: Yes

src/hotspot/share/runtime/continuationFreezeThaw.cpp
robehn: Why do you care about alignment? These are like local variables.

robehn: I removed "int _cont_size"

robehn:
Update, sorry rebased.

src/hotspot/share/runtime/continuationFreezeThaw.cpp
coleenp: I always care about zeros in memory.

robehn: You shouldn't :)

PR #161, opened 2022-04-22
Update problem lists to include newly created CRs

plummercj:
Many svc tests were missing CRs. New CRs have been created and the problem lists updated to reference them.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/161/head:pull/161` \
`$ git checkout pull/161`

`$ git checkout pull/161` \
`$ git pull https://git.openjdk.java.net/loom pull/161/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 161`

`$ git pr show -t 161`

</details>

<a href="https://git.openjdk.java.net/loom/pull/161.diff">https://git.openjdk.java.net/loom/pull/161.diff</a>

</details>

### DIFF
--- a/test/hotspot/jtreg/ProblemList-svc-vthread.txt
+++ b/test/hotspot/jtreg/ProblemList-svc-vthread.txt
@@ -109,37 +109,37 @@ vmTestbase/nsk/jdi/ExceptionEvent/catchLocation/location002/TestDescription.java
 ## to support vthreads and expect the OPAQUE_FRAME error. The others were
 ## not because they don't add any additional value.
 
-vmTestbase/nsk/jdi/Scenarios/invokeMethod/popframes001/TestDescription.java
-vmTestbase/nsk/jdi/BScenarios/hotswap/tc01x002/TestDescription.java
-vmTestbase/nsk/jdi/BScenarios/hotswap/tc02x001/TestDescription.java
-vmTestbase/nsk/jdi/BScenarios/hotswap/tc02x002/TestDescription.java
-vmTestbase/nsk/jdi/BScenarios/hotswap/tc04x001/TestDescription.java
-vmTestbase/nsk/jdi/BScenarios/hotswap/tc04x002/TestDescription.java
-vmTestbase/nsk/jdi/BScenarios/hotswap/tc06x001/TestDescription.java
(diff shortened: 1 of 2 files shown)


AlanBateman:
I think we should add a comment to the top of both files to explain that they are used when running tests with -Dmain.wrapper=Virtual and maybe a sentence to say what that means. Doesn't have to be this PR but looking at these "special" exclude lists makes me think that they will be confusing to find in the repo.

plummercj:
> I think we should add a comment to the top of both files to explain that they are used when running tests with -Dmain.wrapper=Virtual and maybe a sentence to say what that means. Doesn't have to be this PR but looking at these "special" exclude lists makes me think that they will be confusing to find in the repo.

Yes, I agree. In fact this file has tests that shouldn't be in it. For example, the management tests I don't think have any special logic for running in vthread mode, other than jtreg wrapper, which isn't what this problem list is for. They belong in the vthread problem list. I think maybe the jvmti tests belong there too. This problem list is only for tests that fail when not using the jtreg vthread wrapper, but still using `-Dmain.wrapper=Virtual` to put the test in "vthread mode". Explaining all this can get messy, since you need to differentiate between having the main test thread run as a vthread (which is what the jtreg wrapper is for), vs having other threads run in vthread mode by having test specific (or test library) support for creating threads as vthreads.

AlanBateman:
> Yes, I agree. In fact this file has tests that shouldn't be in it. For example, the management tests I don't think have any special logic for running in vthread mode, other than jtreg wrapper, which isn't what this problem list is for. They belong in the vthread problem list. I think maybe the jvmti tests belong there too. This problem list is only for tests that fail when not using the jtreg vthread wrapper, but still using `-Dmain.wrapper=Virtual` to put the test in "vthread mode". Explaining all this can get messy, since you need to differentiate between having the main test thread run as a vthread (which is what the jtreg wrapper is for), vs having other threads run in vthread mode by having test specific (or test library) support for creating threads as vthreads.

Yes, it's complicated to explain but it doesn't need to method jtreg main wrapper or run-test  JTREG_MAIN_WRAPPER=Virtual as it's just for use with serviceability tests that are run with -Dmain.wrapper=Virtual, an execution mode supported by those tests. So I think we can come up with something.

PR #166, opened 2022-04-25
Support yield on virtual thread on EPollSelector select()

joeyleeeeeee97:
Proposal: suport selector poll in loom
Hi, supporting selector poll in loom could enhance compatibility and improve performance based on my experiments with spring  benchmarks.
Many frameworks are using reactor-based networking, for those blocking `Selector.select()` calls currently loom's implementation is blocking the current ForkJoin worker until these calls are returned. For two reasons current virtual threads can't yield out, first is Selector is using synchronized, second is epoll_wait must fall into the kernel. By replacing synchronized with java locks and delegating blocking `Selector.select()` to background threads, this patch supports yielding on `EpollSelector's epoll_wait`. (Similar solutions for coroutine's `Selector.select()` have been widely verified on the production environment in Alibaba).
Hopefully, my benchmarks could be easily reproduced and modified for verification, for spring I have to hack the Thread.start() to start those bundled workers as virtual threads.

In summary, my observations are supporting coroutine yield on virtual threads `Selector.select()` could boost performance when
1. Selector threads(in charge of Selector.select()) are likely to block
2. there are a large number of selector threads 

Numbers:
What did each column mean?
● spring(no suffix) -> the default platform thread mode
● spring-virtualbiz -> all executor threads are converted to virtual threads
● spring-virtual -> all IO threads and executor threads are converted to virtual threads
● spring-virtualoptimized -> all IO threads and executor threads are converted to virtual threads and enable this patch

Spring: 

Configuration:
1. CPU number of Selector threads, each request's handler will be submitted to an executor pool. 
2. Using "wrk" to add pressure, measure latencyAvg and totalRequests each run.
3. Skipped database-related workloads for spring, currently,  they might introduce frequent pinning due to synchronized block.

Summary: 
Converting all executor threads to virtual threads gains more performance than the default, but degradation happens when all threads including IO threads are converted to virtual threads. 

For total requests(throughput) in fixed time, this patch brings **3-7%** improvement on plaintext and **6-14%** improvement on JSON. For average latency, optimize **4-8%** on the plaintext and about 15% on JSON. 

1. We are measuring requests and latency at the same time, so for each run latency and throughput both improved.
2. The percentage is computed via comparing optimized version with current optimal mode(virtualBiz)
```
+------------------------------------------------------------------------------------+
|                       Type: plaintext, Result: totalRequests                       |
+----------+---------+-------------------+-----------------+-------------------------+
| pipeline |  spring | spring-virtualbiz | spring-virtual  | spring-virtualoptimized |
+----------+---------+-------------------+-----------------+-------------------------+
|    4     | 1105401 |      1146776      |     1219481     |         1233398         |
|    8     | 1304546 |      1313621      |     1234331     |         1406274         |
|    16    | 1465894 |      1474328      |     1245648     |         1520790         |
+----------+---------+-------------------+-----------------+-------------------------+
+--------------------------------------------------------------------------------------------+
|                             Type: json, Result: totalRequests                              |
+------------------+---------+-------------------+-----------------+-------------------------+
| concurrencyLevel |  spring | spring-virtualbiz | spring-virtual  | spring-virtualoptimized |
+------------------+---------+-------------------+-----------------+-------------------------+
|        4         |  592597 |       655380      |      698312     |          622674         |
|        8         |  906588 |      1014115      |     1060094     |         1016332         |
|        16        | 1051602 |      1050818      |     1079971     |         1140643         |
|        32        | 1116081 |      1088898      |     1109657     |         1280262         |
|        64        | 1184019 |      1148817      |     1096397     |         1360887         |
|       128        | 1293541 |      1358471      |     1119223     |         1440368         |
+------------------+---------+-------------------+-----------------+-------------------------+
```

How to reproduce:
Environment
Two 4C8G ECS machines on the cloud serve as the java application while the other is in charge of the wrk pressure client.
Based on the [upstream](https://github.com/TechEmpower/FrameworkBenchmarks) Frameworkbenchmark I did some work to adapt this to the loom, now on [the loom](https://github.com/joeyleeeeeee97/FrameworkBenchmarks/) branch we could easily set up a benchmark run to verify this.
● Quickstart( runs on a docker-machine)

```
git clone https://github.com/joeyleeeeeee97/FrameworkBenchmarks.git
cd FrameworkBenchmarks
./tfb --test spring-optimized --duration 60
```

Notice
● if there are socket errors in wrk, then maybe current stress level is too much
● I was using separate machines as client and server following this [guide](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Benchmarking-Getting-Started)
● Use configuration to adjust test duration and concurrency level
--concurrency-levels=[4, 8, 16, 32, 64, 128] --pipeline-concurrency-levels=[128, 256, 512, 1024] --duration 30
Modifications
● Most of my modifications are open now except for the process of building docker image and the hacking part to run spring on loom

Tests
● jtreg java/nio/channels/Selector/ to make sure refactoring Selector synchronized didn't break tests
● jtreg test/jdk/java/lang/Thread/virtual/Selectors.java to verify Selector behavior on virtual threads

```
    @Test
    public void testSelectorMounted() throws Exception {
        var selectorThread =  Thread.ofVirtual().start(() -> {
            try {
                Selector selector = Selector.open();
                selector.select();
            } catch (Exception ignored) {
            }
        });
        Thread.sleep(200);
        // virtual threads are waiting now :)
        assertEquals(selectorThread.getState(),
                (Boolean.parseBoolean(System.getProperty("jdk.useRecursivePoll"))? Thread.State.WAITING : Thread.State.RUNNABLE));
        selectorThread.interrupt();
        selectorThread.join();
    }
```

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/166/head:pull/166` \
`$ git checkout pull/166`

`$ git checkout pull/166` \
`$ git pull https://git.openjdk.java.net/loom pull/166/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 166`

`$ git pr show -t 166`

</details>

<a href="https://git.openjdk.java.net/loom/pull/166.diff">https://git.openjdk.java.net/loom/pull/166.diff</a>

</details>

### DIFF
--- a/src/java.base/linux/classes/sun/nio/ch/EPollSelectorImpl.java
+++ b/src/java.base/linux/classes/sun/nio/ch/EPollSelectorImpl.java
@@ -116,11 +116,18 @@ class EPollSelectorImpl extends SelectorImpl {
 
             do {
                 long startTime = timedPoll ? System.nanoTime() : 0;
-                long comp = Blocker.begin(blocking);
-                try {
-                    numEntries = EPoll.wait(epfd, pollArrayAddress, NUM_EPOLLEVENTS, to);
-                } finally {
-                    Blocker.end(comp);
+                if (Poller.useRecursivePoll() && Thread.currentThread().isVirtual()) {
+                    if (timeout != 0) {
+                        Poller.poll(epfd, Net.POLLIN, TimeUnit.MILLISECONDS.toNanos(timeout), this::isOpen);
+                    }
+                    numEntries = EPoll.wait(epfd, pollArrayAddress, NUM_EPOLLEVENTS, 0);
+                } else {
+                    long comp = Blocker.begin();
+                    try {
+                        numEntries = EPoll.wait(epfd, pollArrayAddress, NUM_EPOLLEVENTS, to);
+                    } finally {
+                        Blocker.end(comp);
+                    }
                 }
                 if (numEntries == IOStatus.INTERRUPTED && timedPoll) {
                     // timed poll interrupted so need to adjust timeout
--- a/src/java.base/share/classes/sun/nio/ch/Poller.java
+++ b/src/java.base/share/classes/sun/nio/ch/Poller.java
@@ -48,6 +48,7 @@ public abstract class Poller {
     private static final Poller[] WRITE_POLLERS;
     private static final int READ_MASK, WRITE_MASK;
     private static final boolean USE_DIRECT_REGISTER;
+    private static final boolean USE_RECURSIVE_POLL;
 
     // true if this is a poller for reading, false for writing
     private final boolean read;
@@ -388,6 +389,12 @@ public abstract class Poller {
         } else {
             USE_DIRECT_REGISTER = "".equals(s) || Boolean.parseBoolean(s);
         }
+        s = GetPropertyAction.privilegedGetProperty("jdk.useRecursivePoll");
+        if (s == null) {
+            USE_RECURSIVE_POLL = provider.useRecursivePoll();
+        } else {
+            USE_RECURSIVE_POLL = "".equals(s) || Boolean.parseBoolean(s);
+        }
         try {
             Poller[] readPollers = createReadPollers(provider);
             READ_POLLERS = readPollers;
@@ -465,5 +472,9 @@ public abstract class Poller {
 
     private Stream<Thread> registeredThreads() {
         return map.values().stream();
+    }
+
+    static boolean useRecursivePoll() {
+        return USE_RECURSIVE_POLL;
     }
 }
--- a/src/java.base/share/classes/sun/nio/ch/PollerProvider.java
+++ b/src/java.base/share/classes/sun/nio/ch/PollerProvider.java
@@ -41,6 +41,10 @@ abstract class PollerProvider {
         return false;
     }
 
+    boolean useRecursivePoll() {
+        return false;
+    }
+
     /**
      * Creates a Poller for read ops.
      */
--- a/src/java.base/share/classes/sun/nio/ch/SelectorImpl.java
+++ b/src/java.base/share/classes/sun/nio/ch/SelectorImpl.java
@@ -41,6 +41,7 @@ import java.util.Iterator;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
 
 
@@ -66,6 +67,9 @@ public abstract class SelectorImpl
 
     // used to check for reentrancy
     private boolean inSelect;
+
+    private final ReentrantLock selectorLock = new ReentrantLock();
+    private final ReentrantLock publicSelectedKeysLock = new ReentrantLock();
 
     protected SelectorImpl(SelectorProvider sp) {
         super(sp);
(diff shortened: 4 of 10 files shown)


AlanBateman:
Selection operations are specified to synchronize on the Selector so we can't change the locking without incompatible changes to the specification. So yes, selection operation on a virtual threads pin the carrier. It is something that can be re-examined once restriction on Java monitors is addressed.

joeyleeeeeee97:
@AlanBateman , Thanks for the tip.  
Does loom consider providing the ability to customize the 'Poller' threads?  For example Netty is using [native implementation](https://netty.io/wiki/native-transports.html) .
If loom provides the ability for users to decide where to register blocking operation and how to cooperate with scheduler, frameworks developers could build up their `IO Poller`. 
For example if loom could opens ability to register blocking event to Poller, and let user start their customized poller as IO mutiplexer.

pron:
Virtual threads are designed to [help application throughput by allowing thousands of threads](https://openjdk.java.net/jeps/425) (or more). Since you're interested in polling on virtual threads, could you explain the use-case for I/O polling on thousands of threads?

AlanBateman:
> @AlanBateman , Thanks for the tip.
> Does loom consider providing the ability to customize the 'Poller' threads? For example Netty is using [native implementation](https://netty.io/wiki/native-transports.html) .
> If loom provides the ability for users to decide where to register blocking operation and how to cooperate with scheduler, frameworks developers could build up their `IO Poller`.
> For example if loom could opens ability to register blocking event to Poller, and let user start their customized poller as IO mutiplexer.

Integration at that level is highly problematic because it is platform specific and FileDescriptor internals cannot be exposed. This means the integration point for frameworks like this is LockSupport.park/unpark. They do their own equivalent of registering for I/O events, use LockSupport.park to park the virtual thread, and then do the wakeup with unpark.

joeyleeeeeee97:
@pron Thanks for your attention.
I assume the optimal solution with IO in loom is using thousands of threads with blocking IO operation, and another layer of I/O polling upon default `Read/Write Poller` is theoretically redundant. But current frameworks need time to evolve.

The main reason for supporting polling is `compatibility` for reactor based current frameworks and users.

For example without this support users are required to update their network framework to a `loom targeted version`(if that work has been done) to achieve the best performance(**The optimized column in benchmark results**), and what if their production is required to work on lower version JDKs?  In that case, they have to maintain a `loom branch` for best performance.

Overall, I think this will make users benefit more loom easier. 

> ```
> +------------------------------------------------------------------------------------+
> |                       Type: plaintext, Result: totalRequests                       |
> +----------+---------+-------------------+-----------------+-------------------------+
> | pipeline |  spring | spring-virtualbiz | spring-virtual  | spring-virtualoptimized |
> +----------+---------+-------------------+-----------------+-------------------------+
> |    4     | 1105401 |      1146776      |     1219481     |         1233398         |
> |    8     | 1304546 |      1313621      |     1234331     |         1406274         |
> |    16    | 1465894 |      1474328      |     1245648     |         1520790         |
> +----------+---------+-------------------+-----------------+-------------------------+
> ```

joeyleeeeeee97:
@AlanBateman . Yes, this patch could be implemented outside the JDK(by framework developers)..

PR #170, opened 2022-04-25
Renaming cleanup on slow path

pchilano:
Some renaming on the slow path. Testing loom-tiers1-3.

---------
### Progress

### Reviewers
 * [Frederic Parain](https://openjdk.java.net/census#fparain) (@fparain - Committer) ⚠️ Review applies to 9bf30f5b33a368d880f6b1ab98867faaa1dffd10

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/loom pull/170/head:pull/170` \
`$ git checkout pull/170`

`$ git checkout pull/170` \
`$ git pull https://git.openjdk.java.net/loom pull/170/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 170`

`$ git pr show -t 170`

</details>

<a href="https://git.openjdk.java.net/loom/pull/170.diff">https://git.openjdk.java.net/loom/pull/170.diff</a>

</details>


src/hotspot/share/runtime/continuationFreezeThaw.cpp
```diff
@@ -389,14 +389,14 @@ class FreezeBase : public StackObj {
   // slow path
   frame freeze_start_frame();
   frame freeze_start_frame_safepoint_stub(frame f);
-  NOINLINE freeze_result freeze(frame& f, frame& caller, int callee_argsize, bool callee_interpreted, bool top);
+  NOINLINE freeze_result recurse_freeze(frame& f, frame& caller, int callee_argsize, bool callee_interpreted, bool top);
```
coleenp, line 392: But now there are two functions named recurse_freeze aren't there?  Same with recurse_thaw?

coleenp: Never mind. I thought I saw an extra recurse_freeze.  The name recurse_thaw is better than thaw_one_frame.

coleenp (approved):

src/hotspot/share/runtime/continuationFreezeThaw.cpp
pchilano: There is only one recurse_freeze. But there are a couple of methods with the recurse_ prefix because they are all called recursively during freeze/thaw. freeze() and thaw() were missing that prefix.

src/hotspot/share/runtime/continuationFreezeThaw.cpp
```diff
@@ -1988,34 +1991,34 @@ void ThawBase::recurse_thaw_compiled_frame(const frame& hf, frame& caller, int n
...
 
-  const int added_argsize = (bottom || caller.is_interpreted_frame()) ? hf.compiled_frame_stack_argsize() : 0;
+  const int added_argsize = (is_bottom_frame || caller.is_interpreted_frame()) ? hf.compiled_frame_stack_argsize() : 0;
   int fsize = ContinuationHelper::CompiledFrame::size(hf) + added_argsize;
   assert(fsize <= (int)(caller.unextended_sp() - f.unextended_sp()), "");
 
-  intptr_t* from = heap_sp - frame::metadata_words;
-  intptr_t* to   = frame_sp - frame::metadata_words;
+  intptr_t* from = heap_top - frame::metadata_words;
+  intptr_t* to   = frame_top - frame::metadata_words;
```
fparain, line 2013: Some names could be more explicit.
For instance, instead of frame_top and heap_top, using stack_frame_top and heap_frame_top would be more precise, making explicit that the top value refers to a frame, which is either still on the stack, or being stored in the heap.

src/hotspot/share/runtime/continuationFreezeThaw.cpp
```diff
@@ -1991,53 +1994,53 @@ NOINLINE void ThawBase::recurse_thaw_interpreted_frame(const frame& hf, frame& c
...
-  intptr_t* const heap_sp = hf.unextended_sp();
-  intptr_t* const frame_bottom = ContinuationHelper::InterpretedFrame::frame_bottom(f);
+  intptr_t* const stack_frame_top = f.sp();
+  intptr_t* const stack_frame_bottom = ContinuationHelper::InterpretedFrame::frame_bottom(f);
+  intptr_t* const heap_frame_top = hf.unextended_sp();
+  intptr_t* const heap_frame_bottom = ContinuationHelper::InterpretedFrame::frame_bottom(hf);
 
   assert(hf.is_heap_frame(), "should be");
-  const int fsize = ContinuationHelper::InterpretedFrame::frame_bottom(hf) - heap_sp;
+  const int fsize = heap_frame_bottom - heap_frame_top;
```
coleenp, line 2011: I think fsize could use a better name too.  Maybe next go-around?

coleenp (approved):
Still good.

pchilano:
Thanks for the reviews @coleenp and @fparain!

PR #178, opened 2022-04-29
8276713

plummercj:
I removed some classpath changes that were causing some JVMTI to fail to find the .smap file for a couple of stratum tests. The changes appear to be unecessary.

I restored the commented out of the code to what it looks like in jdk. I did this to minimize diff with jdk, but if you wish I could either completely remove the commented code, or at least improve the formatting.

Tested with loom-tier1-5 on fibers branch and tier1-5 on jep-vt branch.


AlanBateman (approved):
Thank you for tracking down this difficult to diagnose issue and for testing it on both branches.

PR #191, opened 2022-10-18
Support Kona Fiber of jdk20

miao-zheng:
# Introduce
Kona Fiber is a java stackful fiber implementation in KonaJDK8/11 and has been deployed in a lot java services in Tenent. It sync with most key Loom APIs on early JKU code base. Our team (Tencent Kona JDK) have ported Kona Fiber into OpenJDK20, it passes most Loom test and shows different performance with Loom. We'd pleasure to share Kona Fiber with Loom developers and users as a Loom compitable(mostly) java fiber implementation."

In our port, new option "-XX:+/-UseKonaFiber" is added to turn on Kona Fiber, default is off.  Kona Fiber has better performance on continuation switch but larger memory consumption compared with Loom,  in most scenarios developer can switch from Loom to Kona Fiber without code change.

Below, we elaborate key features on Kona Fiber, its difference with JKU and Loom and share some performance result.

# 1. differences with JKU
Kona Fiber is developed based on early JKU code but we added some key new features and optimization including:
## 1.1 decouple carrier thread and coroutine
JKU coroutine can only run on the Java Thread where coroutine is created. With Loom API(default scheduler is ForkJoinPool), Coroutine or Continuation can run on arbitrary worker thread by scheduler. Kona Fiber decouples coroutine and its creating thread by
1. move coroutine from thread local coroutine list to a global coroutine table.
2. save and restore carrier thread when swtiching between thread and coroutine.

## 1.2 simplify coroutine data structure
JKU allows coroutine switch with monitor\JNI frame, the correctness is guaranteed by developers and JKU only print a warning.So, JKU coroutine keeps local resource area/JNI local handle area/Metadata handle area/MonitorChunk. With Loom restricition, continuation with monitor\JNI frame is not allowed switch. These extra corotuine local data structure is removed, Kona Fiber refactor JKU by
1. refactor inital coroutine start logical, cleanup all handles  and resource areas usage
2. add verification code that checks resoruce area/JNI handle/metadata doesn't change when switch back from coroutine to thread

## 1.3 GC thread parallel walk stack of virtual thread
JKU iterate coroutines when processing its belonging threads during GC or other JVMTI actions, this might causes long GC pause due to unbalanced coroutine creations in JKU. After move coroutines into global table, coroutines is visited parallelly in table buckets.

## 1.4 performance optimization
There are many performance optimizations comparing with JKU, for example: coroutine batch created, cache and reuse stack of coroutine, create coroutine on Demand and so on.

# 2. Adapting Kona Fiber with Loom
## 2.1 runtime modification
Kona Fiber is almostly an independent module. We change the hook of freeze and thaw. Additionally, we implement the stack walking of virtual thread when GC happens. We only achieve the stack walking of G1GC now.

## 2.2 core library modification
In core library, add an extra yield while continuation finish. Because of Kona Fiber is a stackful scheme, we explicit release the stack memory of continuation.

# 3. what Kona Fiber not supported comparing to Loom

### 3.1.1 Continuation not finish but collected by GC
In this scene, Loom also has some problems(see discussion:https://mail.openjdk.org/pipermail/loom-dev/2020-May/001323.html). The extra problem of Kona Fiber is native stack of coroutine not released.

### 3.1.2 ContinuationScope
Kona Fiber do not support scope for continuation hierarchy now, we can support it in future. Continuation can only switch between carrier thread and continuation. 

## 3.2 Switch from carrier thread to virtual thread while carrier thread is pinned
Pin is checked at doYield(switch from virtual thread to carrier thread), but not checked at enterSpecial(switch from carrier thread to virtual thread) on Loom. Kona Fiber adds a dummy continuation for each thread, switch from vt to thread and thread to vt is actually equivalent to Kona Fiber. Virtual thread of Loom can reenter the object monitor which is hold by carrier thread, and Kona Fiber is not because of lightweight object monitor which record address on stack. So Kona Fiber must check Pin when switch from carrier thread to virtual thread, it is different with Loom. 

# 4. Performance result
## 4.1 Continuation switch
Mainly use JMH benchmark at test/micro/org/openjdk/bench/loom/

### 4.1.1 OneShot

The result of OneShot is shown as below while paramCount is 1. 

OneShot.yield: An operation of OneShot.yield is "creat a continuation, run the continuation, and continuation yield". The continuation is created but not finished. We add a cont.run() to make sure continuation is finished, so Kona Fiber can get continuation from cache which can accelerate the process of continuation create.

| stackdepth | 5 | 10 | 20 | 100 |
| ------------ | ------------ | ------------ | ------------ | ------------ | 
| Kona Fiber | 248 | 261 | 278 | 411 |
| Loom | 1508 | 1559 | 1586 | 2379 |

OneShot.yieldAfterEachCall
| stackdepth | 5 | 10 | 20 | 100 |
| ------------ | ------------ | ------------ | ------------ | ------------ | 
| Kona Fiber | 534 | 927 | 1638 | 7441 |
| Loom | 2213 | 3193 | 5237 | 25600 |

OneShot.yieldBeforeAndAfterEachCall
| stackdepth | 5 | 10 | 20 | 100 |
| ------------ | ------------ | ------------ | ------------ | ------------ | 
| Kona Fiber | 903 | 1636 | 3065 | 14581 |
| Loom | 2494 | 3495 | 6062 | 39543 |

OneShot.yieldBeforeEachCall
| stackdepth | 5 | 10 | 20 | 100 |
| ------------ | ------------ | ------------ | ------------ | ------------ | 
| Kona Fiber | 540 | 904 | 1650 | 7547 |
| Loom | 2213 | 3193 | 5237 | 25600 |
 
OneShot.yieldThenContinue
| stackdepth | 5 | 10 | 20 | 100 |
| ------------ | ------------ | ------------ | ------------ | ------------ | 
| Kona Fiber | 284 | 261 | 287 | 405 |
| Loom | 1495 | 1552 | 1589 | 2358 |

### 4.1.2 FreezeAndThaw

The result is yieldAndContinuation minus baseline of testcase FreezeAndThaw. Kona Fiber is 4.x faster than Loom, and Kona Fiber works better when stackdepth get deeper(while stackdepth is 100, Kona Fiber is 20.x faster than Loom), unit is "ns/op":

| stackdepth | 5 | 10 | 20 | 100 |
| ------------ | ------------ | ------------ | ------------ | ------------ | 
| Kona Fiber | 68 | 78 | 56 | 87 |
| Loom | 221 | 269 | 372 | 1807 |

### 4.1.3 Oscillation
Kona Fiber is 2.x to 3.x better than Loom, because there is many time spend on jump from maxDepth to minDepth and the cost is same with Kona Fiber and Loom.(unis is "ns/op")
|  maxDepth | minDepth | Kona Fiber | Loom |
| ------------ | ------------ | ------------ | ------------ |
| 5| 2 | 1012.891 | 3319.44 |
| 5|2|  8047.345 |  22573.425|
|5| 2|  78043.954|  198971.856 |
|5| 3|  1002.794|   3286.773 |
|5| 3|  8095.574|   21109.516|
|5| 3|  77219.734|  196529.532|
|5| 4|  1010.236|   3203.197|
|5| 4|  7661.433|   20111.7|
|5| 4|  73664.17|   187357.237|
|6| 2|  1089.909|   3524.884|
|6| 2|  8515.355|   23636.167|
|6| 2|  83007.198|  223553.905|
|6| 3|  1022.743|   3489.051|
|6| 3|  8222.142|   22415.682|
|6| 3|  79560.057|  215778.073|
|6| 4|  989.495|    3309.326|
|6| 4|  7907.928|   21463.64|
|6| 4|  76620.636|  199450.446|
|7| 2|  1085.62|    3552.785|
|7| 2|  8822.602|   24165.294|
|7| 2|  84875.676|  226516.057|
|7| 3|  1050.816|   3538.169|
|7| 3|  8469.229|   23692.779|
|7| 3|  82485.214|  222592.842|
|7| 4|  1010.377|   3339.384|
|7| 4|  7999.824|   21347.933|
|7| 4|  77563.254|  204326.762|
|8| 2|  1077.988|   3701.969|
|8| 2|  8822.24 |24493.97|
|8| 2|  87558.258|  225938.926|
|8| 3|  1076.211|   3580.486|
|8| 3|  8804.987|   24073.394|
|8| 3|  84496.946|  225824.735|
|8| 4|  1120.996|   3594.21|
|8| 4|  8488.365|   23860.013|
|8| 4|  82291.311|  222835.707|

## 4.2 Continuation create
We have two new JMH benchmark about Continuation create（The benchmark can be seen in the PR）.First case is create and terminate a single coro repeatedly, Second case is batch create 1000 coros and terminate 1000 coros repeatedly. The performance result of Loom with two testcases is similar. Kona Fiber is better at test case 1, and worse at test case 2。

### 4.2.1 single continuation create repeatedly
unit is op/s（means how many continuations are created in one second）
|loom   |kona|
| ------------ | ------------ | 
|2727383|   10658633|

### 4.2.2 batch continuation(batch is 1000)

|loom   |kona|
| ------------ | ------------ | 
|2444375|   1272231|

## 4.3 Memory usage
We create one million virtual threads(each virtual thread execute simple function), and monitor memory usage of the progress. The result is shown below:

|   |java heap|native memory|
| ------------ | ------------ | ------------ | 
|loom|  833M | 177M |
|kona|  275M | 8725M |

### DIFF
--- a/make/autoconf/jvm-features.m4
+++ b/make/autoconf/jvm-features.m4
@@ -46,7 +46,7 @@ m4_define(jvm_features_valid, m4_normalize( \
     \
     cds compiler1 compiler2 dtrace epsilongc g1gc jfr jni-check \
     jvmci jvmti link-time-opt management minimal opt-size parallelgc \
-    serialgc services shenandoahgc static-build vm-structs zero zgc \
+    serialgc services shenandoahgc static-build vm-structs zero zgc fiber \
 ))
 
 # Deprecated JVM features (these are ignored, but with a warning)
@@ -319,6 +319,34 @@ AC_DEFUN_ONCE([JVM_FEATURES_CHECK_STATIC_BUILD],
   ])
 ])
 
+################################################################################
+# Allow to disable fiber
+#
+AC_DEFUN_ONCE([HOTSPOT_ENABLE_DISABLE_FIBER],
+[
+  AC_ARG_ENABLE([fiber], [AS_HELP_STRING([--enable-fiber@<:@=yes/no/auto@:>@],
+      [Default is auto, where fiber is enabled if supported on the platform.])])
+
+  if test "x$enable_fiber" = "x" || test "x$enable_fiber" = "xauto"; then
+    ENABLE_FIBER="true"
+  elif test "x$enable_fiber" = "xyes"; then
+    ENABLE_FIBER="true"
+  elif test "x$enable_fiber" = "xno"; then
+    ENABLE_FIBER="false"
+  else
+    AC_MSG_ERROR([Invalid value for --enable-fiber: $enable_fiber])
+  fi
+
+  if test "x$OPENJDK_TARGET_CPU" != "xx86_64" && test "x$OPENJDK_TARGET_CPU" != "xaarch64"; then
+    ENABLE_FIBER="false"
+    if test "x$enable_fiber" = "xyes"; then
+      AC_MSG_ERROR([Fiber is currently not supported on none X64 platform. Remove --enable-fiber.])
+    fi
+  fi
+
+  AC_SUBST(ENABLE_FIBER)
+])
+
 ###############################################################################
 # Check if the feature 'zgc' is available on this platform.
 #
@@ -391,7 +419,7 @@ AC_DEFUN_ONCE([JVM_FEATURES_PREPARE_PLATFORM],
   JVM_FEATURES_CHECK_SHENANDOAHGC
   JVM_FEATURES_CHECK_STATIC_BUILD
   JVM_FEATURES_CHECK_ZGC
-
+  JVM_FEATURES_CHECK_FIBER
 ])
 
 ###############################################################################
@@ -424,7 +452,7 @@ AC_DEFUN([JVM_FEATURES_PREPARE_VARIANT],
   elif test "x$variant" = "xminimal"; then
     JVM_FEATURES_VARIANT_FILTER="cds compiler2 dtrace epsilongc g1gc \
         jfr jni-check jvmci jvmti management parallelgc services \
-        shenandoahgc vm-structs zgc"
+        shenandoahgc vm-structs zgc fiber"
     if test "x$OPENJDK_TARGET_CPU" = xarm ; then
       JVM_FEATURES_VARIANT_FILTER="$JVM_FEATURES_VARIANT_FILTER opt-size"
     else
--- a/make/hotspot/lib/JvmFeatures.gmk
+++ b/make/hotspot/lib/JvmFeatures.gmk
@@ -162,6 +162,10 @@ ifneq ($(call check-jvm-feature, jfr), true)
   JVM_EXCLUDE_FILES += compilerEvent.cpp
 endif
 
+ifneq ($(call check-jvm-feature, fiber), true)
+  JVM_CFLAGS_FEATURES += -DINCLUDE_KONA_FIBER=0
+endif
+
 ################################################################################
 
 ifeq ($(call check-jvm-feature, link-time-opt), true)
--- a/src/hotspot/cpu/x86/frame_x86.inline.hpp
+++ b/src/hotspot/cpu/x86/frame_x86.inline.hpp
@@ -361,7 +361,7 @@ inline const ImmutableOopMap* frame::get_oop_map() const {
 inline frame frame::sender(RegisterMap* map) const {
   frame result = sender_raw(map);
 
-  if (map->process_frames() && !map->in_cont()) {
+  if (map->process_frames() && !map->in_cont() && !UseKonaFiber) {
     StackWatermarkSet::on_iteration(map->thread(), result);
   }
 
--- a/src/hotspot/cpu/x86/macroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/macroAssembler_x86.cpp
@@ -53,6 +53,9 @@
 #include "runtime/stubRoutines.hpp"
 #include "utilities/macros.hpp"
 #include "crc32c.h"
+#if INCLUDE_KONA_FIBER
+#include "runtime/coroutine.hpp"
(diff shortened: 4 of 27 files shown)


pron:
Hi.
I have a couple of questions.

1. How are continuation stacks resized?
2. Do you have any examples of workloads for which virtual threads are designed -- I/O-heavy concurrent server workloads -- where you see a significant positive performance impact?

-- Ron

miao-zheng:
Hi.
Thanks for your response.

1. We do not support continuation stacks resized until now. In the future, We may map a large virtual space for each continuation stack and uncommit physical memory of continuation stack if it is not used for a long time.
2. Our internal services are implemented on JDK8/11, we have not compared the performance with loom on actual services.

pron:
Thank you.

Given that we cannot maintain _two_ implementations of this functionality (esp. as it is young and likely to have more features added to it), adopting an implementation without a significant real-world benefit and seems to trade off footprint for context-switching latency -- that, while appropriate for other uses of continuations, is the opposite from the tradeoff we need for virtual threads, as [their throughput benefit stems from their lightweight size much more than from fast context-switching](https://inside.java/2020/08/07/loom-performance/) -- I think we'll have to give this one a pass, at least until some benefit that is more pertinent to the intended use-cases can be demonstrated.

-- Ron

miao-zheng:
Thanks for your response.

1. I agree that we should consider this PR after having a significant real-world benefit and new features have been added to Loom.
2. Main purpose of this PR is showing the performance of stackful scheme comparing with Loom. We learn a lot from Loom that the design of virtual thread and corresponding API, and develop a stackful fiber based on early JKU code in JDK8/11 for our internal services.
3. We(Tencent JDK team) will keep paying attention to Loom, adapting new features of Loom into Kona Fiber and try to contribute to Loom.
4. We will continue comparing Kona Fiber with Loom in many scenes, observing the performance in real-world.

PR #193, opened 2022-10-28
ScopedValue javadoc cleanup

AlanBateman:
In process

### DIFF
--- a/src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
+++ b/src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
@@ -30,7 +30,6 @@ import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.lang.ref.Reference;
 import java.util.concurrent.Callable;
-import java.util.function.Function;
 import java.util.function.Supplier;
 import jdk.internal.access.JavaLangAccess;
 import jdk.internal.access.JavaUtilConcurrentTLRAccess;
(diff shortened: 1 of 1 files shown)


src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
```diff
@@ -43,71 +42,109 @@
...
 /**
- * Represents a value that is local to a <em>scope</em>. It is a per-thread value
- * that allows context to be set in a caller and read by callees. The <em>scope</em> is
- * the set of methods that the caller directly invokes, and any methods invoked
- * transitively. Scoped values also provide a way to share immutable data across
- * threads.
+ * A value that is set once and is then available for reading for a bounded period of
+ * execution by a thread. A {@code ScopedValue} allows for safely and efficiently sharing
+ * data for a bounded period of execution without using method arguments.
  *
```
theRealAph, line 48: This reads oddly?
```suggestion
 * data for a bounded period of execution without using passing them as method arguments.
 *
```

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
```diff
@@ -43,71 +42,109 @@
...
- * executed by the {@code run} and {@code call} methods use the {@link #get()} method to
- * read the value of a bound `ScopedValue`. A scoped value reverts to being
- * unbound (or its previous value) when the operation completes.
+ * <p> {@code ScopedValue} defines the {@link #where(ScopedValue, Object, Runnable)}
+ * method to set the value of a {@code ScopedValue} for the period of execution by a
+ * thread of the runnable's {@link Runnable#run() run} method. The unfolding execution of
+ * the methods executed by {@code run} defines a <b><em>dynamic scope</em></b>. The scoped
+ * value is {@linkplain #isBound() bound} when executing in the dynamic scope, it reverts
+ * to being <em>unbound</em> when the {@code run} method completes (normally or with an
+ * exception). Code execting in the dynamic scope uses the {@code ScopeValue} {@link
```
theRealAph, line 55:
```suggestion
 * exception). Code executing in the dynamic scope uses the {@code ScopeValue} {@link
```

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
```diff
@@ -43,71 +42,109 @@
...
+ * method that invokes {@code doSomething()}.
+ * {@snippet lang=java :
+ *     // @link substring="newInstance" target="#newInstance" :
+ *     private static final ScopeValue<String> USERNAME = ScopeValue.newInstance();
+ *
+ *     ScopedValue.where(USERNAME, "duke", () -> doSomething());
+ * }
+ * Code executed directly or indirectly by {@code doSomething()} that invokes {@code
+ * USERNAME.get()} will read the value "{@code duke}". The scoped
+ * value is bound when executing {@code doSomething()} and becomes unbound when {@code
```
theRealAph, line 73:
```suggestion
 * value is bound while executing {@code doSomething()} and becomes unbound when {@code
```

?

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
```diff
@@ -43,71 +42,109 @@
...
- * scoped value already has a value, then {@code run} or {@code call} methods may be
- * invoked to run another operation with a newly-bound value. Code executed by the
- * operation will read the new value of the scoped value. The scoped value reverts to its
- * previous value when the operation completes.
+ * <p> Consider the following example with a scoped value {@code USERNAME} that is
+ * <em>bound</em> to the value "{@code duke}" for the execution, by a thread, of a run
+ * method that invokes {@code doSomething()}.
+ * {@snippet lang=java :
+ *     // @link substring="newInstance" target="#newInstance" :
+ *     private static final ScopeValue<String> USERNAME = ScopeValue.newInstance();
```
theRealAph, line 67:
```suggestion
 *     private static final ScopedValue<String> USERNAME = ScopeValue.newInstance();
```

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
theRealAph: `s/ScopeValue/ScopedValue/g`

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
```diff
@@ -43,71 +42,109 @@
...
+ * bind or read its value.
  *
- * Scoped-value bindings can be shared across threads when used in conjunction with
- * {@link StructuredTaskScope}. Creating a {@code StructuredTaskScope} captures the
- * current thread's scoped-value bindings for inheritance by threads {@link
- * StructuredTaskScope#fork(Callable) forked} in the task scope. This means that a thread
- * may bind a scoped value and share its value in a structured concurrency
- * context. Threads forked in the task scope that read the scoped value will read
- * the value bound by the thread that created the task scope.
+ * <p> The value of a {@code ScopedValue} should be an immutable object. This is
```
theRealAph, line 90: This statement seems a bit strong. A large part of the usages need the value of a `ScopedValue` to be mutable: object pools, output streams, sockets...

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
```diff
@@ -43,71 +42,109 @@
...
+ *     ScopedValue.where(USERNAME, "duchess", () -> doMore());
+ * }
+ * Code executed directly or indirectly by {@code doMore()} that invokes {@code
+ * USERNAME.get()} will read the value "{@code duchess}". When {@code doMore()} completes
+ * (normally or with an exception), the value of {@code USERNAME} reverts to
+ * "{@code duke}".
+ *
+ * <h2><a id="inheritence">Inheritance</a></h2>
+ *
+ * {@code ScopedValue} supports sharing data across threads. At this time, this sharing
```
theRealAph, line 115:
```suggestion
 * {@code ScopedValue} supports sharing data across threads. This sharing
```

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
```diff
@@ -43,71 +42,109 @@
...
+ *             scope.fork(() -> childTask3());
+ *
+ *             ...
+ *          }
+ *     });
  * }
  *
+ * <p> Unless otherwise specified, passing a {@code null} argument to a method in this
+ * class will cause a {@link NullPointerException} to be thrown.
+ *
```
theRealAph, line 145: Should this last sentence be elsewhere? It's rather lost at the end of the "Inheritance" section.

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
AlanBateman: Yes, and "passing the data" might be a better again.

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
AlanBateman: Fair point, better to drop it to avoid any confusion.

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
AlanBateman: The insertion of the Rebinding and Inheritance headings causds it to get pushed down, making it look like its in the Inheritance section.

theRealAph (approved):
I think that's it for now. I have a few more changes to make outside the region you touched.

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
```diff
@@ -43,71 +42,105 @@
...
- *   // @link substring="newInstance" target="ScopedValue#newInstance" :
- *   private static final ScopedValue<Credentials> CREDENTIALS = ScopedValue.newInstance();
+ *     ScopedValue.where(USERNAME, "duchess", () -> doMore());
+ * }
+ * Code executed directly or indirectly by {@code doMore()} that invokes {@code
+ * USERNAME.get()} will read the value "{@code duchess}". When {@code doMore()} completes
+ * (normally or with an exception), the value of {@code USERNAME} reverts to
+ * "{@code duke}".
+ *
+ * <h2><a id="inheritence">Inheritance</a></h2>
```
theRealAph, line 112:
```suggestion
 * <h2><a id="inheritance">Inheritance</a></h2>
```

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
```diff
@@ -220,10 +253,30 @@ static final Snapshot getInstance() {
...
-     * An immutable map of scoped values to values.
-     * It define the {@link #run(Runnable) run} and {@link #call(Callable) call} methods
-     * to invoke an operation with the scoped value mappings bound to the thread
-     * that invokes {@code run} or {@code call}.
+     * A mapping of scoped values, as <em>keys</em>, to values.
+     *
+     * <p> A {@code Carrier} is used to accumlate mappings so that an operation (a
+     * {@link Runnable} or {@link Callable}) can be executed with all scoped values in
+     * the mapping bound to a value. The following example runs an operation with k1
+     * bound (or rebound) to v1, and k2 bound (or rebound) to v2.
```
theRealAph, line 261:
```suggestion
     * {@link Runnable} or {@link Callable}) can be executed with all scoped values in
     * the mapping bound to values. The following example runs an operation with k1
     * bound (or rebound) to v1, and k2 bound (or rebound) to v2.
```

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
```diff
@@ -220,10 +253,30 @@ static final Snapshot getInstance() {
...
+     *     ScopedValue.where(k1, v1).where(k2, v2).run(() -> ... );
+     * }
+     * It is equivalent to:
+     * {@snippet lang=java :
+     *     // @link substring="where" target="ScopedValue#where(ScopedValue, Object, Runnable)" :
+     *     ScopedValue.where(k1, v1, () -> {
+     *         ScopedValue.where(k2, v2, () -> ... );
+     *     });
+     * }
+     *
```
theRealAph, line 273: This is a bit circular: we define
```
ScopedValue.where(key, value, op)
``` 
as equivalent to
```
ScopedValue.where(key, value).run(op);
```
then we define 
```
ScopedValue.where(k1, v1).where(k2, v2).run(op);
```
as equivalent to
```
ScopedValue.where(k1, v1, () -> {
        ScopedValue.where(k2, v2, () -> ... );
    });
```

so we're defining them in terms of each other. I guess that's OK.

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
```diff
@@ -43,71 +42,105 @@
...
+ * class will cause a {@link NullPointerException} to be thrown.
  *
- * <p> Unless otherwise specified, passing a {@code null} argument to a constructor
- * or method in this class will cause a {@link NullPointerException} to be thrown.
+ * <h2><a id="rebind">Rebinding</a></h2>
  *
- * @apiNote
- * The following example uses an scoped value to make credentials available to callees.
+ * The {@code ScopedValue} API allows a new binding to be established for <em>nested
+ * dynamic scopes</em>. This is known as <em>rebinding</em>. A {@code ScopedValue} that
```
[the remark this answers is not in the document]
AlanBateman: I just reading this paragraph and I think it is okay as is, only because the lead up does not use the word "operation" or "action.

src/jdk.incubator.concurrent/share/classes/jdk/incubator/concurrent/ScopedValue.java
AlanBateman: The intention here was to show that the chaining avoids needing to nest the lambda expressions. I'll make to make this better to avoid the circularity.

PR #202, opened 2023-09-01
Add a new subclass of StructuredTaskScope that shows the finished subtasks as a Stream

forax:
This is a minimal patch that adds a new subclass of StructuredTaskScope named Stremable (better name needed) pushing failed/succceding subtasks into a Stream.

This subclass aim to:
- make easier for users to use STS without having to override handleCompleted, which is called concurrently so hard to get right, at a price of being a little less efficient
- ease the implementation of shortcuited stream semantics like get the first two values, get the first value greater than a threshold, etc by auto shutdowning the STS once the condition is true

The Streamable STS adds two new methods joinWhile/joinUntilWhile(function) that takes a function that takes a Stream and return a value
```java
  public <U> U joinWhile(Function<? super Stream<Subtask<T>>, ? extends U> mapper) throws InterruptedException {
```
When this method is called, each finished subtask (with state SUCCESS or FAILED) are pushed into the Stream until there is no more subtasks, the stream has finished (has been short-circuited), the scope has been shutdown, interrupted or the dealine occurs. If some tasks are still pending because the stream has been short-cirtuited, they are shutdown.

Here are two examples:
- get a list of all the values that suceeed
```java
    try(var streamable = new StructuredTaskScope.Streamable<Integer>()) {
            streamable.fork(() -> {
                Thread.sleep(200);
                return 12;
            });
            streamable.fork(() -> {
                Thread.sleep(100);
                return 17;
            });
            List<Integer> list = streamable.joinWhile(stream -> stream.filter(task -> task.state() == State.SUCCESS).map(Subtask::get).toList());
            System.out.println(list);  // [17, 12]
        }
```
- find the first subtask (that suceed or fail)
```java
    try(var streamable = new StructuredTaskScope.Streamable<Integer>()) {
            streamable.fork(() -> {
                Thread.sleep(1_000);
                return 12;
            });
            streamable.fork(() -> {
                Thread.sleep(100);
                return 17;
            });
            Optional<Subtask<Integer>> first = streamable.joinWhile(Stream::findFirst);
            System.out.println(first);  // Optional[PlainSubTask[state=SUCCESS, result=17, exception=null]]
        }
```
Internally, handleCompleted post each subtask into a queue which is read by the Stream spliterator inside joinWhile.

The current implementation uses thread flock methods ThreadFlock.awaitAll()/ThreadFlock.wakeup().
If we want the implementation to only depends on protected methods, still not sure if its a good idea, we need
- expose ensureOwner and ensureOpen,
- expose round values (forkRound, lastJoinAttempted, lastJoinCompleted), this can be done using a method that takes a lambda
- expose ThreadFlock.awaitAll() ThreadFlock.wakeup().

---------
### Progress

### Reviewers
 * @donPain (no known openjdk.org user name / role) ⚠️ Review applies to [6cdf85ae](https://git.openjdk.org/loom/pull/202/files/6cdf85ae8cebc0843d4276024f948f09d0d1d48d)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/loom.git pull/202/head:pull/202` \
`$ git checkout pull/202`

`$ git checkout pull/202` \
`$ git pull https://git.openjdk.org/loom.git pull/202/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 202`

`$ git pr show -t 202`

</details>

<a href="https://git.openjdk.org/loom/pull/202.diff">https://git.openjdk.org/loom/pull/202.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/loom/pull/202#issuecomment-1702743502)

### DIFF
--- a/src/java.base/share/classes/java/util/concurrent/StructuredTaskScope.java
+++ b/src/java.base/share/classes/java/util/concurrent/StructuredTaskScope.java
@@ -330,7 +330,7 @@ public class StructuredTaskScope<T> implements AutoCloseable {
      * @since 21
      */
     @PreviewFeature(feature = PreviewFeature.Feature.STRUCTURED_CONCURRENCY)
-    public sealed interface Subtask<T> extends Supplier<T> permits SubtaskImpl, Streamable.PlainSubTask {
+    public sealed interface Subtask<T> extends Supplier<T> permits SubtaskImpl, Streamable.PlainSubtask {
         /**
          * {@return the value returning task provided to the {@code fork} method}
          *
@@ -1352,7 +1352,9 @@ public class StructuredTaskScope<T> implements AutoCloseable {
      */
     @PreviewFeature(feature = PreviewFeature.Feature.STRUCTURED_CONCURRENCY)
     public static final class Streamable<T> extends StructuredTaskScope<T> {
-        private record PlainSubTask<T>(State state, T result, Throwable exception) implements Subtask<T> {
+        private record PlainSubtask<T>(State state, T result, Throwable exception) implements Subtask<T> {
+            private final static PlainSubtask<?> STOP = new PlainSubtask<>(null, null, null);
+
             @Override
             public Callable<? extends T> task() {
                 throw new UnsupportedOperationException();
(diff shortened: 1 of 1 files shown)


forax:
I've updated the implementation to use ThreadFlock.wakeup()/ThreadFlock.awaitAll() so the semantics of joinWhen() is now fully compatible with join()/joinUntil().
- you can call joinWhen() several times,
- you can call shutdown() when you want,
- a subtask can spawn another subtask,
- etc

I've not exposed all the methods I need as protected method because I'm still not convince that the implemntation of joinWhen() is not too specific. And I still need to add a variant of joinWhen() that takes a timeout as parameter.

As a bonus example, here is an example that use patiningBy to store all failed tasks in a List and all tasks that suceed in another list
```java
try(var streamable = new StructuredTaskScope.Streamable<Integer>()) {
     streamable.fork(() -> { throw new IOException("oops"); });
     streamable.fork(() -> 17);
            
     Map<Boolean, List<Subtask<Integer>>> map =
         streamable.joinWhen(stream -> stream.collect(partitioningBy(task -> task.state() == State.SUCCESS)));
     System.out.println(map);
         // {false=[PlainSubTask[state=FAILED, result=null, exception=java.io.IOException: oops]],
         //   true=[PlainSubTask[state=SUCCESS, result=17, exception=null]]}
}
```

donPain (approved):

forax:
keep it alive

PR #205, opened 2023-09-08
modify VirtualThreadPinnedEventThrows.java

ywj0101:
This test is for testing parking when pinned and VirtualThreadPinnedEvent.commit fails with OOME.However, this test can also pass if OOM is not thrown.Therefore, it is not possible to determine whether VirtualThreadPinnedEvent has thrown an exception.

In reference to the parkOnCarrierThread method in VirtualThread in JDK 22, if OOM (Out of Memory Error) is thrown, then the event will be null.
```java
private void parkOnCarrierThread(boolean timed, long nanos) {
        assert state() == RUNNING;

        VirtualThreadPinnedEvent event;
        try {
            event = new VirtualThreadPinnedEvent();
            event.begin();
        } catch (OutOfMemoryError e) {
            event = null;
        }
```
So, I added a condition to confirm that OOM (Out of Memory Error) has been thrown and that the event is null (assertEquals(0, virtualThreadPinnedEventCount);).

### DIFF
--- a/test/jdk/java/lang/Thread/virtual/VirtualThreadPinnedEventThrows.java
+++ b/test/jdk/java/lang/Thread/virtual/VirtualThreadPinnedEventThrows.java
@@ -29,6 +29,20 @@
  * @run junit VirtualThreadPinnedEventThrows
  */
 
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.locks.LockSupport;
+import java.util.stream.Collectors;
+import java.lang.management.ManagementFactory;
+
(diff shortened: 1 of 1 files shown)


AlanBateman:
The purpose of this test is to exercise code paths for parking while pinned that could potentially throw OOME. It's essentially that park does not fail. It's hard to create reliable tests for OOME so this test runs with a patched version of VirtualThreadPinnedEvent that allows the test control if OOME thrown by any of the 3 methods where it interacts with JFR (create, begin, commit). So JFR is not used. If I understand your change correctly, you want this extended to test that a JFR event is not recorded. Do you view this as a "test for the test"? I'm just wondering if this additional complexity (and overlap with the JfrEvents test) is needed.

PR #206, opened 2023-10-12
C2 changes to support Java Object Monitors with Interpreter

ahmedmuhsin:
Only known issue is with the ProfileInterpreter on however, we implemented a work around: 9bca6603204e94ffac4be450b73610cbaf9eed0f 

Testing:
Compared jtreg results from jom-master to our PR with ProfileInterpreter off and with ProfileInterpreter on with the work around. Note that with ProfileInterpreter on and no work around jtreg errors out (so that configuration is not included in the test tables below).

See test results in comments.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/loom.git pull/206/head:pull/206` \
`$ git checkout pull/206`

`$ git checkout pull/206` \
`$ git pull https://git.openjdk.org/loom.git pull/206/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 206`

`$ git pr show -t 206`

</details>

<a href="https://git.openjdk.org/loom/pull/206.diff">https://git.openjdk.org/loom/pull/206.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/loom/pull/206#issuecomment-1759944223)


macarte:
@pchilano / @dean-long - ready to be sponsered

src/hotspot/share/opto/graphKit.cpp
```diff
@@ -3507,6 +3507,22 @@ void GraphKit::shared_unlock(Node* box, Node* obj) {
...
   unlock->init_req(TypeFunc::Parms + 0, obj);
   unlock->init_req(TypeFunc::Parms + 1, box);
+
+  if (ObjectMonitorMode::java()) {
+
+#ifdef ASSERT
+    // For some reason, we sometimes end up with dead locals that must be killed
+    // before we call add_safepoint_edges.
+    if (!dead_locals_are_killed()) {
+      kill_dead_locals();
```
dean-long, line 3517: It's because some callers of shared_unlock() are not calling kill_dead_locals() first.  Instead of calling shared_unlock() directly, callers could push the object and then call do_monitor_entery() instead.

src/hotspot/share/opto/parse1.cpp
```diff
@@ -1118,11 +1119,26 @@ void Parse::do_exits() {
...
+
       if (do_synch) {
 #ifdef C2_PATCH
         if (ObjectMonitorMode::legacy()) {
 #endif
+        // Ensure that the jvms object has a method pointer inorder to
+        // avoid asserts.
+        if (ObjectMonitorMode::java()) {
+          if (!ex_jvms->has_method()){
+            ex_jvms->set_method(method());
```
dean-long, line 1131: We probably don't need this if we find out why the caller state doesn't have a method.  I'll see if I can find the problem.

PR #210, opened 2024-07-19
test PR

dougxc:
PR to make it easier to see diff to master using the GitHub PR UI.

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [ ] Change must not contain extraneous whitespace
- [ ] Commit message must refer to an issue

### Integration blockers
&nbsp;⚠️ Whitespace errors (failed with updated jcheck configuration in pull request)
&nbsp;⚠️ Whitespace errors (failed with updated jcheck configuration in pull request)
&nbsp;⚠️ Whitespace errors (failed with updated jcheck configuration in pull request)
&nbsp;⚠️ Whitespace errors (failed with updated jcheck configuration in pull request)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/loom.git pull/210/head:pull/210` \
`$ git checkout pull/210`

`$ git checkout pull/210` \
`$ git pull https://git.openjdk.org/loom.git pull/210/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 210`

`$ git pr show -t 210`

</details>

<a href="https://git.openjdk.org/loom/pull/210.diff">https://git.openjdk.org/loom/pull/210.diff</a>

</details>


src/hotspot/cpu/x86/c2_CodeStubs_x86.cpp
```diff
@@ -73,32 +73,59 @@ void C2EntryBarrierStub::emit(C2_MacroAssembler& masm) {
...
+  // C2CodeStubList::emit() will throw an assertion and report the actual size that
+  // is needed.
+  return DEBUG_ONLY(40) NOT_DEBUG(25);
+}
+
+void C2HandleAnonOMOwnerStub::emit(C2_MacroAssembler& masm) {
+  __ bind(entry());
+  Register mon = monitor();
+  Register t = tmp();
+  __ movptr(t, Address(r15_thread, JavaThread::lock_id_offset()));
```
mur47x111, line 88: This stub is removed at https://github.com/openjdk/jdk/commit/7f6bb71eb302e8388c959bdaa914b758a766d299,  and I believe it was lightweight locking specific before the removal. It is now only used by stack locking in the fibers branch, which I believe is not necessary @AlanBateman

pchilano: The reason why we need to identify this case is to decrement _held_monitor_count. For LM_LEGACY I changed the code to only track stack-locked acquired monitors (and only first time, not for the recursive case). This means if we would go to the inflated path the counter will not be decremented. Once we use tid as owner for inflated monitors we don't need to track them anymore since there is no issue with those during freeze. The stack-lock ones on the other hand are still problematic and we need to fix them during freeze. The _held_monitor_count should be called _monitors_to_fix_on_freeze. Now, we are likely going to remove the loom monitor support for LM_LEGACY, i.e we will still pin with monitors. In that case I think this can go away since we can go back to tracking all monitors.

PR #225, opened 2026-06-15
Route all VT creation through scheduler.newThread() for custom schedulers

franz1981:
When a custom scheduler is installed via implClass, all VT creation routes through scheduler.newThread(String, int, Thread, Runnable) — the new raw-params overload. 
The existing builder overload delegates to it, extracting name and characteristics from the builder.

Single routing point in newVirtualThread(): when scheduler is null and CUSTOM_SCHEDULER is true, routes through the default scheduler's raw newThread(). No cycle — the raw default creates VirtualThread directly.

For paths without a builder (startVirtualThread, factory), uses no builder allocation — raw params are passed directly.

---------

This is a PoC which show 2 things:
1. if the (custom) scheduler overrides the `newThread` method can wrap its `Runnable` with a `ScopedValue` which can help with #224: still duplication (as it is now, which is not great), but enable **every VT** to run with such context.
2. using the `preferredCarrier != null` info, a custom scheduler can "treat" differently (by forcing some specific scheduling policy) such `VirtualThread` e.g. JDK pollers

Other 2 things:
- I dislike to pass/intercepts raw VT parameters
- Since we have some "legacy" thread builder `scheduler` field, this looks more messy than it could, sorry for that: with the unified "custom scheduler" API probably there's no reason to have per-builder `scheduler` fields

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/loom.git pull/225/head:pull/225` \
`$ git checkout pull/225`

`$ git checkout pull/225` \
`$ git pull https://git.openjdk.org/loom.git pull/225/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 225`

`$ git pr show -t 225`

</details>

<a href="https://git.openjdk.org/loom/pull/225.diff">https://git.openjdk.org/loom/pull/225.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/Thread.java
+++ b/src/java.base/share/classes/java/lang/Thread.java
@@ -968,25 +968,40 @@ public class Thread implements Runnable {
          * @implSpec The default implementation creates a new virtual thread. It should
          * be rare to override this method.
          *
-         * @param builder the virtual thread builder
-         * @param preferredCarrier the preferred carrirer, can be {@code null}
+         * @param name the thread name, can be {@code null}
+         * @param characteristics the thread characteristics
+         * @param preferredCarrier the preferred carrier, can be {@code null}
          * @param task the object to run when the thread executes
          * @return the {@code VirtualThreadTask} that scheduler executes
          * @throws UnsupportedOperationException if this is the built-in default scheduler
+         */
+        default VirtualThreadTask newThread(String name, int characteristics,
+                                            Thread preferredCarrier,
+                                            Runnable task) {
+            Objects.requireNonNull(task);
+            if (this == VirtualThread.builtinScheduler(false)) {
+                throw new UnsupportedOperationException();
+            }
+            var vthread = new VirtualThread(this, preferredCarrier, name, characteristics, task);
+            return vthread.virtualThreadTask();
+        }
+
+        /**
+         * Creates a new virtual thread using the given builder's configuration.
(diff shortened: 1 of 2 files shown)


src/java.base/share/classes/java/lang/Thread.java
```diff
@@ -968,27 +968,42 @@ public interface VirtualThreadScheduler {
...
+         * @param characteristics the thread characteristics
+         * @param preferredCarrier the preferred carrier, can be {@code null}
          * @param task the object to run when the thread executes
          * @return the {@code VirtualThreadTask} that scheduler executes
          * @throws UnsupportedOperationException if this is the built-in default scheduler
-         *
-         * @see <a href="Thread.html#inheritance">Inheritance when creating threads</a>
          */
-        default VirtualThreadTask newThread(Builder.OfVirtual builder,
+        default VirtualThreadTask newThread(String name, int characteristics,
```
AlanBateman, line 978: characteristics is JDK internal implementation default so cannot be exposed with the API like this.

franz1981: I know, I mentioned it in the first desc

> I dislike to pass/intercepts raw VT parameters

That's what I meant :pray: 

This is just a PoC to show the idea: currently there are different routes (builders/factories and raw VT start), and unify them to all go through the scheduler::newThread which uses a builder is not still right

PR #229, opened 2026-09-15
Make Virtual Thread Contructor Public

Muthu-Palaniyappan-OL:
Hey Loom Team,  

You guys are doing amazing work, thank you for your amazing contributions. I have a small suggestion for [VirtualThread.java](https://github.com/openjdk/jdk21u/blob/82c3abe7290356da33078ea53283af18b4157a38/src/java.base/share/classes/java/lang/VirtualThread.java#L152-L181). Please can you guys make this constructor public ? I have presented my case down below. 

**_Problem:_**
* I am developing a multi tenant system where many users can essentially create a lot of virtual threads. 
* I am okay with creating a lot of Virtual Threads and keep it waiting, but i am uncomfortable with spinning a lot of OS Threads. 
* I want to put a restriction on the number of OS threads each user can create. 
Something like this, there can be 10s virtual threads for each user in an application but each user in an app will have only one OS thread. 
* This is currently not possible with existing systems because the only way for me to do it is using Semaphores, but that makes me lose the benefits of virtual threads, i wouldn't be able to multiply a lot of virtual threads in limited OS threads. 

**_Ask:_**
* I would like to make VirtualThread(Executor, String, int, Runnable) public so i can pass custom executor (ForkJoinPool or something)
Reason:
* I wanna be able to "multiplex" a lot of virtual threads on a particular controlled set of OS threads, even if the OS thread is not getting allocated is fine for me. 
* This change makes multi-tenant systems feasible. and virtual threads a scalable solution for all. 

**_Alternatives:_**
* I could possibly use reflection but that's not recommended because of performance reasons. 
Using "jdk.virtualThreadScheduler.maxPoolSize" will not solve the "multi tenant" problem. 

**_My Offer:_**
* If u want i will also be happy to create a PR for you folks from JDK21, JDK25, JDK27, Master. 
Something similar to https://github.com/openjdk/jdk21u/pull/509 
* I want thoughts about this idea from you folks. Thanks for looking into this. This is not a bug / issue. That's why I am sending this email to start some discussion. 

Regards,
Muthu Palaniyappan Olagappan

---------
- [ ] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).

---------
### Progress

### Error
&nbsp;⚠️ Pull request body is missing required line: `- [x] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/loom.git pull/229/head:pull/229` \
`$ git checkout pull/229`

`$ git checkout pull/229` \
`$ git pull https://git.openjdk.org/loom.git pull/229/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 229`

`$ git pr show -t 229`

</details>

<a href="https://git.openjdk.org/loom/pull/229.diff">https://git.openjdk.org/loom/pull/229.diff</a>

</details>


Muthu-Palaniyappan-OL:
Hey @fisk @pchilano , sorry for tagging you 🙏 ... please can i get your opinon on this idea ? am i even doing the right thing ? was this intended ? 

Thanks a lot 🙇  ... (i already signed my OCA... just waiting for approval...)

